Issue #17, opened 2021-07-05
Theme translation messages support


markodvornik:
I would like to add translation files to the theme.

### Expected behaviour

Keycloak [documentation on internationalization](https://www.keycloak.org/docs/latest/server_development/#internationalization) states that in order to support a custom language two steps are neccessary:

1. create a `messages*` file inside the theme folder
2. modify `theme.properties` by adding the `locales= ...` entry

#### (1) Messages

I was investigating if translations are already supported. AFAIK the build theme script `generateKeycloakThemeResources.js` only copies the `resources` folder to the `theme` folder. The `messages` folder is not mentioned in the script.

To be clear what I was looking for: Keycloak [expects](https://www.keycloak.org/docs/latest/server_development/#deploying-themes) the `messages` folder next to the `resources` folder.

```
    META-INF/keycloak-themes.json
    theme/mytheme/login/theme.properties
    theme/mytheme/login/login.ftl
    theme/mytheme/login/resources/css/styles.css
    theme/mytheme/login/resources/img/image.png
    theme/mytheme/login/messages/messages_en.properties
    theme/mytheme/email/messages/messages_en.properties
```

Maybe I am missing something here, so any help regarding adding translation files would be much appreciated.

A workaround would be to add the `messages` folder to the theme by [updating](https://docs.oracle.com/javase/tutorial/deployment/jar/update.html) a `jar` package once it is compiled, but that is not a permanent solution.

#### (2) Modify theme.properties

I think this pull-request solves it: https://github.com/InseeFrLab/keycloakify/pull/18

garronej:
Hi @markodvornik, 
I am sorry I didn't implement any mechanism to update the default message.  
I made the assumption (wrong obviously) that pepoles would want to use their own translation engine.
I currently don't have time to update Keycloakify but what you can do is use the hook.
However you can use the langage hook to tell what language is currently selected. 
```typescript
import {  useKcLanguageTag  } from "keycloakify";

const { kcLanguageTag, setKcLanguageTag } = useKcLanguageTag(); 
```

For example if language selected is Slovenian you will get `kcLanguageTak === "sk"`, if it's english `kcLanguageTak === "en"` after that you are free to display you string it whatever way you see fit.  
[...]

PR #18, opened 2021-07-06
Add possibility to add custom properties to theme.properties file

asashay:
Right now upon generation of the `theme.properties` only `parent=keycloak` is added inside.

In our use case, we need to add some custom value pairs there (e.g. `env=dev`). Added some code to enable it.

Cheers!

### DIFF
--- a/src/bin/build-keycloak-theme/build-keycloak-theme.ts
+++ b/src/bin/build-keycloak-theme/build-keycloak-theme.ts
@@ -19,6 +19,7 @@ export function main() {
     console.log("🔏 Building the keycloak theme...⌚");
 
     const extraPagesId: string[] = (parsedPackageJson as any)["keycloakify"]?.["extraPages"] ?? [];
+    const extraThemeProperties: string[] = (parsedPackageJson as any)["keycloakify"]?.["extraThemeProperties"] ?? [];
 
     generateKeycloakThemeResources({
         keycloakThemeBuildingDirPath,
@@ -55,7 +56,8 @@ export function main() {
             };
 
         })(),
-        extraPagesId
+        extraPagesId,
+        extraThemeProperties
     });
 
     const { jarFilePath } = generateJavaStackFiles({
--- a/src/bin/build-keycloak-theme/generateKeycloakThemeResources.ts
+++ b/src/bin/build-keycloak-theme/generateKeycloakThemeResources.ts
@@ -23,12 +23,13 @@ export function generateKeycloakThemeResources(
         //If urlOrigin is not undefined then it means --externals-assets
         urlOrigin: undefined | string;
         extraPagesId: string[];
+        extraThemeProperties: string[];
     }
 ) {
 
     const { 
         themeName, reactAppBuildDirPath, keycloakThemeBuildingDirPath, 
-        urlPathname, urlOrigin, extraPagesId
+        urlPathname, urlOrigin, extraPagesId, extraThemeProperties
     } = params;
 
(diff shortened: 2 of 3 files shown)


markodvornik:
I wonder if this is related to Issue #17 (Theme translation). 

Keycloak documentation on [internationalization](https://www.keycloak.org/docs/latest/server_development/#internationalization) states that in order to support a custom language two steps are neccessary:

1. create a `messages*` file inside the theme folder
2. modify `theme.properties` with `locales=...` entry

> To add a new language create the file <THEME TYPE>/messages/messages_<LOCALE>.properties in the directory of your theme. Then add it to the locales property in <THEME TYPE>/theme.properties. For a language to be available to users the realms login, account and email theme has to support the language, so you need to add your language for those theme types.

AFAIK, #17 needs this possibility.

garronej:
Hi @asashay,  
Thanks for the PR, it's well implemented and all but I must be missing something because I don't understand why you can't pass your extra properties directly to your react components? 🤔 
There is nothing in the `theme.properties` because everything is [here](https://github.com/InseeFrLab/keycloakify/blob/main/src/lib/components/KcProps.ts) and the `parent=keycloak` is only here so that for the pages that I havent implemented we fallback to the `keyclaok` theme instead of crashing, [ref](https://github.com/InseeFrLab/keycloakify#some-pages-still-have-the-default-theme-why).   

I have an idea that would explain your need to do that...
If you have a `.ftl` pages that you don't want to port to react that you copy manually to `build_keycloak/src/main/resources/theme/[keycloak or <your_app>]/login/` just before the theme is bundled. Is that what you do?

asashay:
Hey @garronej , thanks for the prompt reply.

Having some custom variables sitting in the `theme.properties` will require adding them to all pages contexts (TS at least), so they are accessible everywhere. And it's not very straightforward and error-prone. Is this your concern?

In our case, we are building and using Keycloak in multiple environments - dev, test, production.
And having the possibility to add custom properties will help to reference environment variables with nice fallback as described [here](https://www.keycloak.org/docs/latest/server_development/#:~:text=not%20found%20with%20%24%7B-,foo%3AdefaultValue,-%7D.). IMHO having such option could be helpful in some cases.

Our other option is to add different .env files for different environments though.

Best

garronej:
> Having some custom variables sitting in the `theme.properties` will require adding them to all pages contexts (TS at least), so they are accessible everywhere. And it's not very straightforward and error-prone. Is this your concern?

No, I have no problem with that. It's just that I don't understand why you need to implement it this way.  

How do you actually access that custom property ? By doing `window.kcContext.properties.env` ?  

I assume your code isn't open-source but can you paste here the relevant snippet?

garronej:
I will merge since there is at least one usecase that I understand. 
Still interested to know more about your usecase though.
Thanks for the PR!

asashay:
> How do you actually access that custom property ? By doing `window.kcContext.properties.env` ?

I'm adding it to the context of every page, like:
```
type Env = {
    properties: {
        env: 'dev' | 'development' | 'production'
    }
}

export const { kcContext } = getKcContext<
        {
            pageId: "login.ftl"
        } & Env |
        {
            pageId: "register.ftl"
        } & Env |
        ....
        >()
```

and then accessing it the same way as the rest of variables in the kcContext.

Having the possibility to extend Common context for all pages would be really helpful here. But this is part of a different discussion 🙂

garronej:
Ok thank you for the precisions.  
Do you sometime need to update the env in `theme.properties` or do you build it and never touch it again?

asashay:
we add it once and then do not touch it. `theme.properties` is very rarely modified

garronej:
@all-contributors please add @asashay for tests and code

PR #28, opened 2021-08-04
fix(mvn): scoped packages compatibility

marcmrf:
Pom generated for scoped packages like `@scope/package` is malformed. Artifact id is generated using `package.json` name field, being an invalid pattern.

```
[ERROR] 'artifactId' with value '@scope/package' does not match a valid id pattern.
```

This PR sanitizes package name for using `scope-package` instead.


garronej:
Thank you for contributing and well done.

garronej:
I published a new version with your fix.

garronej:
@all-contributors please add @marcmrf for tests and code

PR #63, opened 2021-12-08
fix: errors in common.ftl

Ann2827:
ref: https://github.com/InseeFrLab/keycloakify/issues/58

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,5 @@ jspm_packages
 
 /sample_react_project/
 /.yarn_home/
+
+.idea
--- a/src/bin/build-keycloak-theme/generateFtl/common.ftl
+++ b/src/bin/build-keycloak-theme/generateFtl/common.ftl
@@ -1,7 +1,6 @@
-<script>const _= 
-<#macro objectToJson_please_ignore_errors object depth>
+<script>const _=
+<#macro objectToJson_please_ignore_errors object depth arr ftl>
     <@compress>
-
         <#local isHash = false>
         <#attempt>
             <#local isHash = object?is_hash || object?is_hash_ex>
@@ -11,9 +10,7 @@
             <#return>
         </#attempt>
         <#if isHash>
-
             <#local keys = "">
-
             <#attempt>
                 <#local keys = object?keys>
             <#recover>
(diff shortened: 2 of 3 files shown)


marziman:
@garronej any chance to get priority for this to be reviewed?

garronej:
@Ann2827 Thank you for your work.  

Sorry for my lack of enthusiasm yesterday, I didn't have time to look closely at what you did. Thank you for your work on Keycloakify, it's very valuable!  
All the more so that since Keycloak 15 the loading time is getting out of hand. We have to, at least, mitigate the problem of `ftl` errors in logs.  

(I am sure that you are aware of the following point that I am going to make but I will make them anyway for the sake of documenting our discussion.) 

I am confident that what you did fixes all the errors, however, I think it needs some more work to be future proof.  
What if, in the next version of Keycloak, there is an extra property in the  `url` object. It won't be listed [here](https://github.com/Ann2827/keycloakify/blob/29fbcdc0a6583f1611ac677113c1653684b7f87e/src/bin/build-keycloak-theme/generateFtl/common.ftl#L44) so it won’t be present in the `kcContext` .   
There are also some people writing Keycloak plugins that add non-standard ftl values. We want to list them as well. For example [this plugin](https://github.com/micedre/keycloak-mail-whitelisting) that creates `kcRegisterContext["authorizedMailDomains"]`. 

As a result I would like to take an alternative approach that would consist of avoiding performing action that we know will throw. 

For example, if we know that dereferencing `.data_model.foo.bar.baz` throws in the template `example.ftl` we could avoid doing it. With this "best effort" approach, we would be sure not to break anything.  

What do you think?

@marziman  

> any chance to get priority  

I don't know, what do you mean by any chance?  
There is still a lot of work to be done on this PR before it gets merged. I maintain a lot of open-source projects so it could be a couple of weeks or more before it's available. 

Best regard.

marziman:
@garronej 
thanks alot for your reply.
I am also maintaining OSS projects and know about that situation as maintainer.

I was thinking you are fine with her changes. Thats why I asked. I believe filling up logs with errors, which are not critucal is bad in many ways. I think that this shall get high priority, cause everyone who uses Keycloakify will feel the same shock that sth. is wrong and see the exceptions.

Please dont take it wrong, I love the project. But this could make people not use your valuable work here.

If we can support let us know.
Maybe we can get a Keycloakify version with @Ann2827 changes to help testing.

I am sure it will need a proper review first.

@garronej I believe that this issue might need much attention 🙏🏻✌🏻

BR Mehmet

marziman:
@Ann2827 
can we test your fixes on our project? Do you have a version where we could grab your changes? Or shall we build Keycloakify from source with your applied changes?

garronej:
@marziman The problem is that the fixes of @Ann2827 are likely to break the login pages of other company that use Keycloak plugins.  

Believe me, I don't like the errors in the logs either but as discussed [here](https://github.com/InseeFrLab/keycloakify/discussions/42#discussioncomment-1494703) there is no way in the FreeMarker language to try catch without login something.  
There is no good solution to this problem, the only think we can do is avoid doing things we know by experience will throw.  

I will publish a beta version incorporating @Ann2827 changes but PR are welcome.

garronej:
@Ann2827 @marziman [Here is](https://www.npmjs.com/package/keycloakify/v/4.2.14-beta.0) the beta version.

garronej:
I close this PR, let's continue [here](https://github.com/InseeFrLab/keycloakify/pull/65)

PR #69, opened 2021-12-21
Adding name field to realm in KcContext type

VBustamante:
This PR adds the `name` field to the realm object in the KcContext Common type that is already being sent to the theme but wasn't declared, causing a compilation error when used.

image

I'm sending this PR directly because the fix seemed easy enough, but please let me know if there's something I misunderstood or missed


garronej:

garronej:
I relased a new build incorporing your change.

VBustamante:
Thank you for the amazing project!

garronej:
Thank you!

PR #70, opened 2021-12-21
Added realm name field to KcContext mocks object

VBustamante:
PR https://github.com/InseeFrLab/keycloakify/pull/69 broke build because KcContext mock didn't create the realm name field. This PR fixes the issue.

Sorry about that 😰


garronej:
No problem!

Issue #102, opened 2022-05-05
Add support for 2FA setup page


tylucaskelley:
I was working on customizing the theme for the `LoginOtp` page, and found that the other side of the 2FA support - the initial setup page (screenshot below) doesn't seem to be supported in this library. Is there a plan to add support for it?

![image](https://user-images.githubusercontent.com/5448759/167020394-fdd7a920-cb10-43a1-aa32-02aa935120bf.png)

garronej:
Hi, @tylucaskelley 

I just added [a documentation section](https://docs.keycloakify.dev/limitations#some-pages-still-have-the-default-theme.-why) for you.  

Best regard

PR #127, opened 2022-06-28
feat: add login-config-totp.ftl page

aidangilmore:
Adds `login-config-totp.ftl` page, resolving #102

### DIFF
--- a/src/bin/build-keycloak-theme/generateFtl/generateFtl.ts
+++ b/src/bin/build-keycloak-theme/generateFtl/generateFtl.ts
@@ -20,6 +20,7 @@ export const pageIds = [
     "login-idp-link-confirm.ftl",
     "login-idp-link-email.ftl",
     "login-page-expired.ftl",
+    "login-config-totp.ftl",
 ] as const;
 
 export type PageId = typeof pageIds[number];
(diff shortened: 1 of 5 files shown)


garronej:
Wonderfull, thank you very much, looks clean.

garronej:
@aidangilmore,  
Yes just one maybe, have you tested in Keycloak or just with `yarn start`?  
Have you linked your version of Keycloakify in `keycloakify-demo-app`?  
I'm asking because I just created the doc, I would like to know how it went for you.

garronej:
I released a new version including your changes.

garronej:
Yes, thank you for the lovely feedback.  

> didn’t have any major issue following the contributing document, just a small one caused by the versions of React not 
> lining up between Keycloakify devDependencies and keycloakify-demo-app.   

It makes sense that you had issues linking react, I forgot, Keycloakify is still using 17 when the demo app is using 18.  I'll correct this.  

> It wasn't entirely clear how to get the types for KcContextBase, so I just looked at the Keycloak source code and used 
> The types from there. If there is a better way of doing this, it may be worth mentioning in the doc.  

No, unfortunately there is no better way than to infer from usage just like you did.  

> I will say that I really liked that you pointed to a sample commit that I could base my changes off of.  

OK good to know!  

Thanks again for taking the time to contribute to Keycloakify!

garronej:
@all-contributors please add @aidangilmore for tests and code

PR #160, opened 2022-08-18
Avoiding superfluous downloads by not deleting the temporary keycloak download folder

francoisautin:
Hi !

I have been developing for a little while a user friendly keycloak theming service using your library. I've been really happy with it so far!

An issue I've found though, is that keycloakify downloads the same file over and over while it could be kept in cache. This simple commit avoids downloading the keycloak archive if it already exists. If it isn't found (for example because of a version bump in the new keycloakify version), the folder is deleted and the new archive is downloaded.

This increases the speed of the generation process and avoids failure from when github rejects repeated requests.

There could be optimization still, for instance by not unziping the archive, but this simple fix is already ample enough.

### DIFF
--- a/src/bin/tools/downloadAndUnzip.ts
+++ b/src/bin/tools/downloadAndUnzip.ts
@@ -2,7 +2,7 @@ import { basename as pathBasename, join as pathJoin } from "path";
 import { execSync } from "child_process";
 import fs from "fs";
 import { transformCodebase } from "./transformCodebase";
-import { rm_rf, rm, rm_r } from "./rm";
+import { rm_rf } from "./rm";
 
 /** assert url ends with .zip */
 export function downloadAndUnzip(params: { url: string; destDirPath: string; pathOfDirToExtractInArchive?: string }) {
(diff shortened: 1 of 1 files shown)


garronej:
Hi @francoisautin,  
Thank you for the PR!  
Related to https://github.com/InseeFrLab/keycloakify/issues/103

francoisautin:
Fixed a small issue where the temporary folder containing the keycloak zip found itself in the final .jar by adding explicit exclusion of this directory in the pom.xml

garronej:
Hi @francoisautin,  
Thank you very much for the PR truly but I have decided to implement this myself in the end because to do the caching well there are some subtleties.  
For starter we need to make sure that the download was compleate and suscessfull before adding it to the cache also `downloadAndUnzip.ts` should not generate garbage in the directory that will be bundled in the jar.  
I have implemented a cache mechanism in v6, it will be released soon.  

Thanks again for your input, I'll be happy to have your name in the contributor list, if you want to submit any PR, even trivial on the v6 branch I'll be happy to merge.   

Best,

francoisautin:
Thank you very much for your work on this feature! No worries about having my name in the contributor list! I understand that the way I approached this problem was kind of naïve and I would much rather rely on your code than my own :sweat_smile:

PR #166, opened 2022-08-30
docs(README): fix typo

civilizeddev:
While v6 is just around the corner, I think it is still worthwhile to fix the current page.


garronej:
Thanks you!

PR #175, opened 2022-09-20
Fix typo in development docs

PinPinIre:
I noticed this tiny typo when reading through the documentation. 

There also seems to be a typo in the file name `developpement.md`. Should it be `development.md`? I have kept the filename as it is, in case changing would break links.


garronej:
Thank you for your help!

Issue #217, opened 2023-01-04
Possible security flow : username enumeration if anti brute force activated


remigauthierdocaposte:
Hello all,

We noticed that if the brute force is activated and if an account is blocked, the response generated using Keycloakify different than with a classic user/password failure. Thus the attacker will be able to induce that a username exists when blocking the account. If this is automated, an attacker can guess all username and carry on other attacks using these sensitive informations (targetted phishing, password guess...).

The payload incriminated on browser side :

If the user is not blocked :

     out["messagesPerField"]= {
    
        "printIfExists": function (fieldName, x) {
                    if(fieldName === "global" ){
                            return "" ? x : undefined;
                    }
                    if(fieldName === "userLabel" ){
                            return "" ? x : undefined;
                    }
                    if(fieldName === "username" ){
                            return "" ? x : undefined;
                    }
                    if(fieldName === "email" ){
                            return "" ? x : undefined;
                    }
                    if(fieldName === "firstName" ){
                            return "" ? x : undefined;
                    }
                    if(fieldName === "lastName" ){
                            return "" ? x : undefined;
                    }
                    if(fieldName === "password" ){
                            return "1" ? x : undefined;

If the user is blocked :

    out["messagesPerField"]= {
    
        "printIfExists": function (fieldName, x) {
                    if(fieldName === "global" ){
                            return "" ? x : undefined;
                    }
                    if(fieldName === "userLabel" ){
                            return "" ? x : undefined;
                    }
                    if(fieldName === "username" ){
                            return "1" ? x : undefined;
                    }
                    if(fieldName === "email" ){
                            return "" ? x : undefined;
                    }
                    if(fieldName === "firstName" ){
                            return "" ? x : undefined;
                    }
                    if(fieldName === "lastName" ){
                            return "" ? x : undefined;
                    }
                    if(fieldName === "password" ){
                            return "" ? x : undefined;

I am going to see if we can bypass this behavior on our side and if this is a missuse of the lib.
Meanwhile, I was hoping that you could investigate on your side and give a feedback on this.

Thanks in advance,
Rémi

garronej:
Interesting, thank you for investigating this.  
At first thought I'd say that Keycloak should assume that info exposed to FTL isn't sensible.  
Otherwise it means that that theme author can inadvertently introduce security vulnerability.  
If we agree on this premise, I'd say that it's an issue for Keycloak.  
[...]

PR #218, opened 2023-01-06
fix - handle username and password login errors the same way

pedrodsRocha1:
Addresses #217.

As mentioned in the replies, this is most likely a server-side issue that Keycloak should look into themselves.
However, as it poses a significant vulnerability, I've created a quick fix so that people using this library won't be affected by it.

Tested with a custom theme built using the lib, should also be tested using both starter projects.


garronej:
You are the real MVP <3
I'll reveiw tonight.  
It's FTL e.g. it's free solo, I have to focus

PR #221, opened 2023-01-08
fix: replace rm system calls with nodejs native functions; closes #219

### LINKED ISSUE #219: Use of execSync to remove files and dirs makes this nonportable
Unfortunately, we cant run keycloakify from powershell or cmd, because of the use of system calls to execute `rm <args>`. (e.g. https://github.com/InseeFrLab/keycloakify/blob/2bfbba4daf786018de4ec7c519651f20658c0c17/src/bin/tools/rm.ts)
Consider using a nodejs function: https://stackoverflow.com/a/57866165/1568684

lordvlad:
replace rm system calls with nodejs native functions; closes #219


garronej:
Thank you very much for taking care of this!

garronej:
I'm releasing a new version

PR #231, opened 2023-01-23
keycloak test script: use env to launch bash

rome-user:
This minor change allows users to use the latest version of `bash` on Mac OS.

`/bin/bash` is very old on Mac OS, and many users will have `bash` from MacPorts or Homebrew. This change allows such users to be able to use the newer `bash` installation.


garronej:
Nice, thanks!

garronej:
Releasing new version

PR #239, opened 2023-02-03
Mvn begone

### LINKED ISSUE #220: Using maven to bundle the jar add superfluous dependency
Effectively, maven creates a zip file. In addition to the theme files, there are 4 more files, two of which are created by keycloakify itself. We could easily replace the maven dependency by creating the zip archive using nodejs functions.

lordvlad:
will close #220 

This PR aims to add alternatives to using `mvn` to create the final bundled theme jar.

mvn is an additional dependency, which needs to be installed, configured, has its own dependencies, and makes building in a CI (or even windows) more cumbersome than need be.

The `keycloakify` section in `package.json` has now an additional optional option  called `bundler`, which accepts the values `['mvn', 'keycloakify', 'none']`. If nothing is set, it will default to `mvn`. This option can also be overriden using the environment variable `KEYCLOAKIFY_BUNDLER`, accepting the same values of course.

* Choosing `mvn` (or defaulting) will result in an unchanged build
* Choosing `none` will not create a jar. The user will need to call maven or gradle or whatever themselves.
* Choosing `keycloakify` will use the new bundler code to create a jar file using nodejs libs only, no additional system libraries required.

In addition, this PR adds options to explicitely set artifactId using the `keycloakify` section in `package.json`, as well as options to override groupId and artifactId using the environment variables `KEYCLOAKIFY_GROUP_ID` and `KEYCLOAKIFY_ARTIFACT_ID`. The reasoning is, that companies often have naming conventions in place, which can now be upheld more easily.

Also added is an option to override the `version` from `package.json` using the environment variable `KEYCLOAKIFY_VERSION`. In our CI, for example, the version is determined from git tags and I don't want to patch package.json before running keycloakify.


lordvlad:
In a follow-up PR, I intend to make `build_keycloak` directory configurable

src/bin/keycloakify/BuildOptions.ts
```diff
@@ -40,8 +48,9 @@ export namespace BuildOptions {
         themeName: string;
         extraPages?: string[];
         extraThemeProperties?: string[];
-        //NOTE: Only for the pom.xml file, questionable utility...
```
lordvlad, line 43: not 'only' for pom file, but also used when uploading the jar to a package registry, e.g. sonatype nexus or gitlab

src/bin/keycloakify/generateJavaStackFiles.ts
```diff
@@ -16,21 +18,19 @@ export type BuildOptionsLike = {
 }
 
 export function generateJavaStackFiles(params: {
-    version: string;
```
lordvlad, line 19: looked like duplication to me, let me know if I was too eager

src/bin/keycloakify/keycloakify.ts
```diff
@@ -45,17 +46,30 @@ export async function main() {
...
         buildOptions
     });
 
-    child_process.execSync("mvn package", {
-        "cwd": keycloakThemeBuildingDirPath
-    });
+    if (buildOptions.bundler === "none") {
+        logger.log("😱 Skipping bundling step, there will be no jar");
+    } else if (buildOptions.bundler === "keycloakify") {
+        logger.log("🫶 Let keycloakify do its thang");
```
lordvlad, line 57: typo on purpose.

garronej:
Wow, nice 😻   
I'm reviewing this this week end!

src/bin/keycloakify/generateJavaStackFiles.ts
garronej: No you weren't 👍🏻

garronej:
Hi @lordvlad,  
Thank you very much for your work, it's realy great to drop the dependence to Maven and to have alle thoses new build options!   

I documented everything: https://docs.keycloakify.dev/build-options#keycloakify.bundler  

I made some little adjustmenet mostly related to type safery, you can check them out here:  https://github.com/InseeFrLab/keycloakify/pull/240  

The main thing I did is to set `keycloakify` to be the default bundler.  
You did all this great work why not make it the default.  

Again, many thanks,  

Cheers

garronej:
Re @lordvlad,  
Unless it's my mistake again, I can't build with the new `"bundler": "keycloakify"` option.  
When I set the option in the keycloakify-advanced-starter I get this runtime error:  

image

I'm on Mac M1 v16.19.0.  

I've put back "mvn" as the default for now...

lordvlad:
I'll double check

lordvlad:
follow-up in #241

PR #241, opened 2023-02-05
Mvn begone addendum

lordvlad:
So emberassing, introduced an 'improvement' to readability in last second. Unfortunately, nodejs streams are untyped. Will try and improve this further with typed streams in a follow-up PR

### DIFF
--- a/src/bin/tools/jar.ts
+++ b/src/bin/tools/jar.ts
@@ -1,11 +1,12 @@
 import { Readable, Transform } from "stream";
 import { pipeline } from "stream/promises";
-import { relative, sep } from "path";
+import { dirname, relative, sep } from "path";
 import { createWriteStream } from "fs";
 
 import walk from "./walk";
 import type { ZipSource } from "./zip";
 import zip from "./zip";
+import { mkdir } from "fs/promises";
 
 /** Trim leading whitespace from every line */
 const trimIndent = (s: string) => s.replace(/(\n)\s+/g, "$1");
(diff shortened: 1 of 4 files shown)


src/bin/tools/jar.ts
```diff
@@ -57,9 +57,9 @@ export default async function jar({ groupId, artifactId, version, rootPath, targ
     const pathToRecord = () =>
         new Transform({
             objectMode: true,
-            transform: function (path, _, cb) {
-                const filename = relative(rootPath, path).split(sep).join("/");
-                this.push({ filename, path });
+            transform: function (fsPath, _, cb) {
+                const path = relative(rootPath, fsPath).split(sep).join("/");
+                this.push({ path, fsPath });
```
lordvlad, line 62: stronger typing here would've caught the error :(

src/bin/tools/jar.ts
```diff
@@ -69,19 +69,21 @@ export default async function jar({ groupId, artifactId, version, rootPath, targ
             }
         });
 
-    /**
-     * Create an async pipeline, wait until everything is fully processed
-     */
-    await pipeline(
+    await mkdir(dirname(targetPath), { recursive: true });
+
+    // Create an async pipeline, wait until everything is fully processed
```
lordvlad, line 74: Okay  this block is super strange. While double-checking the behavior, I noticed that the final success banner was not printed. I experimented a bit and in fact _nothign_ I tried to print after `await pipeline(...)` ever got printed, though the produced jar file was fine, and there was no error thrown anywhere. Will need to investigate this, but reverting it to a bit more naive implementation using a hand-crafter promise yielded the desired result.

garronej:
Thank you for the quick fix!

PR #263, opened 2023-03-15
Fix start container script paths for windows

mkreuzmayr:
The `generateStartKeycloakTestingContainer.ts` script is corrupt on windows because paths contain backslashes, I added code to replace all backslashes and additionally quoted the paths.


garronej:
Yes nice, thank you very much!

PR #268, opened 2023-03-20
Asynchronize transform and download pipeline

mkreuzmayr:
Asynchronize transform and download pipeline and (nearly) every system call to increase build performance. I also fixed download caching by checking the path, at least on my machine download caching did not work as expected. I managed to get build time down from ~35 seconds to ~13 seconds if downloads are cached on my machine.

### DIFF
--- a/src/bin/keycloakify/keycloakify.ts
+++ b/src/bin/keycloakify/keycloakify.ts
@@ -3,13 +3,14 @@ import { generateJavaStackFiles } from "./generateJavaStackFiles";
 import { join as pathJoin, relative as pathRelative, basename as pathBasename } from "path";
 import * as child_process from "child_process";
 import { generateStartKeycloakTestingContainer } from "./generateStartKeycloakTestingContainer";
-import * as fs from "fs";
+import * as fs from "fs/promises";
 import { readBuildOptions } from "./BuildOptions";
 import { getLogger } from "../tools/logger";
 import { getCliOptions } from "../tools/cliOptions";
 import jar from "../tools/jar";
 import { assert } from "tsafe/assert";
 import type { Equals } from "tsafe";
+import { existsSync } from "fs";
 
 const reactProjectDirPath = process.cwd();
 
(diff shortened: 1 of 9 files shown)


src/bin/keycloakify/keycloakify.ts
garronej, line 1: Not worth it

src/bin/tools/crawl.ts
garronej, line 1: Here there was an oportunity for real optimization. But not like this.  

If we do a for loop with an `await` we sould aswell do it asyncrhonously.  

What would improve the performance is to be able to run the callback on the file as soon as we see it instead to collecting everything then returning the array.

garronej (commented):
First of all, thank you very much for your willingness to contribute and for pointing out that there is room for optimization performances.  
That being said, this PR can't be merged. 
- The DIFF is too big for a single PR. Some changes are indentation related.  
- There are a lot of things that are changed that have virtually no impact on performances and the real opportunities for improvement are not actioned upon.  

Now I would be very willing to merge a promise that would fix the caching issue you are experiencing @lordvlad could you have a look at what's going on?  

Besides, I promise to improve the build performance in the very near future.  
Again thank you for your involvement, if you're still willing to contribute we need help on implementing the Account pages...

mkreuzmayr:
Should I reopen a PR with only parts of it? Or do you have any plans for a different implementation?

mkreuzmayr:
I´ll PR the cache issue tomorrow. I understand why you did not merge it, I just went all in with my local changes.

src/bin/keycloakify/keycloakify.ts
```diff
@@ -136,4 +139,6 @@ export async function main() {
             ""
         ].join("\n")
     );
+
+    console.timeEnd("keycloakify");
```
lordvlad, line 143: should this stay in prod code?

src/bin/tools/crawl.ts
lordvlad: check walk.ts.

src/bin/tools/downloadAndUnzip.ts
```diff
@@ -71,14 +69,24 @@ async function getNpmProxyConfig(): Promise<Pick<FetchOptions, "proxy" | "noProx
...
+        return filepath;
+    }
+
+    const proxyOpts = await getNpmProxyConfig();
+
+    const response = await fetch(url, proxyOpts);
+
     await writeFile(filepath, response.body);
+
+    console.log(`Downloaded ${url} to ${filepath}!`);
```
lordvlad, line 88: should this stay in production code? everywhere else @garronej used custom logger

lordvlad: @garronej would you consider introducing https://www.npmjs.com/package/debug? I've been using it for a long time and its super handy cause you don't need to pass a config from module to module

src/bin/tools/transformCodebase.ts
```diff
@@ -20,27 +20,29 @@ export function transformCodebase(params: { srcDirPath: string; destDirPath: str
         }))
     } = params;
 
-    for (const file_relative_path of crawl(srcDirPath)) {
+    const filePromises = (await crawl(srcDirPath)).map(async file_relative_path => {
```
lordvlad, line 23: maybe you can use `walk.ts`?

`
await Promise.all(Array.from(walk(srcDirPath)).map(async relativeFilePath => {
 ....
})))
`

garronej:
Thank you very much <3

src/bin/tools/transformCodebase.ts
lordvlad:
```suggestion
    const filePromises = (await crawl(srcDirPath)).map(async fileRelativePath => {
```

garronej: To save some real time I'd like an api more like 

```ts
export type Crawl = (onFileFound: (fileRelativePath: string)=> void)=> Promise<void>;
```

It would be faster than having to collect all file to start processing them.

garronej:
@lordvlad Thanks for the exta eyes on this!

lordvlad:
```
$  curl -vvv -O /dev/null https://github.com/keycloak/keycloak/archive/refs/tags/21.0.1.zip
...
< cache-control: max-age=0, private
```

🙄 whoops

lordvlad:
okay since we use stable urls, we can simply do the following

```
    const opts: FetchOptions = { cachePath, cache: 'force-cache', ...proxyOpts };
```

PR #269, opened 2023-03-21
fix(download): fix download cache not behaving as expected

lordvlad:
Should fix the download cache reported by @mkreuzmayr


garronej:
> Wouldn´t it be more appropriate to use node_modules/.cache/keycloakify as a cache folder?  

Yes absolutely, thanks for this suggestion  

> is kinda expensive to run. (4 terminal exec´s at worst case)

Does it have actual impact on performance though. We used to have a custom file based cache mechanism in place, @lordvlad abstracted away this logic. I like it abstracted. How much improvement are we talking about?  
Maybe we should try benchmarking the build time between the current version and the last one with file based caching, see if there is a significant difference. My guesse it there won't be.

mkreuzmayr:
@garronej I like iterating on a development environment as close as possible to the production environment I am deploying to. That´s why I have a docker container running with the build output passed in the container´s theme directory. After making changes I run keycloakify build and the keycloak theme gets updated when the is finished.

Yes I know I could just use the mock feature for that (that´s what I did in the beginning for the basic layout). But for testing out my user profile fields and auth flows I just kept iterating by building the theme and using it in my local keycloak container. For me this was a much better experience than using the mockData, even though the build times sometimes were >40s. I am pretty sure that it is possible to reduce the build time for this workflow way under 10s to make this a much better developer experience.

This is also why I was completely over-engineering the performance improvements in recent PR. I just wanted to squeeze out every millisecond of performance to show what could be possible.

mkreuzmayr:
> Does it have sensible impact on performance though? We used to have a custom file based cache mechanism in place, @lordvlad abstracted away this logic. I like it abstracted.

Don´t understand me wrong, I like abstraction too, but I think for donwloading the file I would just use the on-board nodejs https module and cache the file in `node_modules/.cache` and for unzipping I´d use some tested external library like `yauzl` ([used in vscode](https://github.com/microsoft/vscode/blob/main/package.json#L97)) or `node-stream-zip`.

garronej:
> @garronej I like iterating on a development environment as close as possible to the production environment I am deploying to. That´s why I have a docker container running with the build output passed in the container´s theme directory. After making changes I run keycloakify build and the keycloak theme gets updated when the is finished.
> 
> Yes I know I could just use the mock feature for that (that´s what I did in the beginning for the basic layout). But for testing out my user profile fields and auth flows I just kept iterating by building the theme and using it in my local keycloak container. For me this was a much better experience than using the mockData, even though the build times sometimes were >40s. I am pretty sure that it is possible to reduce the build time for this workflow way under 10s to make this a much better developer experience.
> 
> This is also why I was completely over-engineering the performance improvements in recent PR. I just wanted to squeeze out every millisecond of performance to show what could be possible.

Thanks for the explaination @mkreuzmayr, that makes a lot of sence. Let's squeeze out every millisecond of performance then.  
Already the download and extract are made in paralel like you suggested. Now I think we can win big by doing by doing the crawl and replace in //.

garronej:
> I fixed the npm config calling too many times and caching on a WIP branch, so those Performance considerations should be addressed.
> 
> Note though, that npm config is unrelated to caching.
> 
> The http lib that comes with npm makes proxy handling rather cumbersome, that's why I opted for make-fetch-happen. Unfortunately their cache implementation is useless as GitHub will always return max-age=0 headers, so they will always assume it's stale. I re-introduced custom caching.
> 
> node_modules/.cache vs $XDG_CACHE_HOME/.keycloakify I don't care either way, but node_modules/.cache is definitely a better fallback than home directory, I'll take that into account.

Thank you @lordvlad for all this work and investigation you are doing 😍  
The `max-age=0` is really a bummer it would have been a very elegant solution.  

> node_modules/.cache vs $XDG_CACHE_HOME/.keycloakify I don't care either way, but node_modules/.cache is definitely a better fallback than home directory, I'll take that into account.  

👍🏻

lordvlad:
if you worry about the missing `cacheDirPath` and `isSilent` arguments: cacheDirPath is now obsolete and isSilent will be restored in a later step

garronej:
Hi @lordvlad,  
Thank you very much!  
Sorry I miss your message for some reason,  next time undraft when you think it's ready.  

No need to apologies you're I am very grathfull of **any** time you spend of Keycloakify.

scripts/generate-i18n-messages.ts
```diff
@@ -37,7 +37,9 @@ const logger = getLogger({ isSilent });
         const baseThemeDirPath = pathJoin(tmpDirPath, "base");
 
         crawl(baseThemeDirPath).forEach(filePath => {
-            const match = filePath.match(/^([^/]+)\/messages\/messages_([^.]+)\.properties$/);
+            const match =
+                filePath.match(/^([^/]+)\/messages\/messages_([^.]+)\.properties$/) ||
+                filePath.match(/^([^\\]+)\\messages\\messages_([^.]+)\.properties$/);
```
garronej, line 42: It woule be cleaner to use `new Regexp(`([^\\${path.sep}]+)\/messages\\${path.sep}messages_([^.]+)\.properties$`)`) but okay.

PR #272, opened 2023-03-22
feat(login context): improve login typings

willwill96:
Keycloak only treats `rememberMe` as true if the value is `"on"`. [ref](https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/authentication/authenticators/browser/AbstractUsernameFormAuthenticator.java#L210)

Additionally, it seems that keycloak will also return the attempted password on failed attempts, which could be useful to pre-populate the form on errors:
image

### DIFF
--- a/src/login/kcContext/KcContext.ts
+++ b/src/login/kcContext/KcContext.ts
@@ -101,7 +101,8 @@ export declare namespace KcContext {
         registrationDisabled: boolean;
         login: {
             username?: string;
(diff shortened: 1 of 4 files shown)


garronej:
@willwill96 !!  
Man your rock!!!! 
I was wondering why the remember me checkbox didn't seem to work.  

Thank you so much!

Issue #285, opened 2023-03-29
keycloakify 7.3.1 starter project issues


mkreuzmayr:
I am getting these errors when updating keycloakify to 7.3.1 in the starter project. Any ideas?

```
TypeError: args[i].replace is not a function
    at populateTemplate (D:\Temp\keycloakify-starter\node_modules\keycloakify\bin\tools\trimIndent.js:49:33)
    at trimIndentPrivate (D:\Temp\keycloakify-starter\node_modules\keycloakify\bin\tools\trimIndent.js:61:35)
    at trimIndent (D:\Temp\keycloakify-starter\node_modules\keycloakify\bin\tools\trimIndent.js:84:30)
    at D:\Temp\keycloakify-starter\node_modules\keycloakify\bin\tools\jar.js:71:68
    at step (D:\Temp\keycloakify-starter\node_modules\keycloakify\bin\tools\jar.js:37:23)
    at Object.next (D:\Temp\keycloakify-starter\node_modules\keycloakify\bin\tools\jar.js:18:53)
    at D:\Temp\keycloakify-starter\node_modules\keycloakify\bin\tools\jar.js:12:71
    at new Promise (<anonymous>)
    at __awaiter (D:\Temp\keycloakify-starter\node_modules\keycloakify\bin\tools\jar.js:8:12)
    at jar (D:\Temp\keycloakify-starter\node_modules\keycloakify\bin\tools\jar.js:60:12)
    at D:\Temp\keycloakify-starter\node_modules\keycloakify\bin\keycloakify\keycloakify.js:133:60
    at step (D:\Temp\keycloakify-starter\node_modules\keycloakify\bin\keycloakify\keycloakify.js:56:23)
    at Object.next (D:\Temp\keycloakify-starter\node_modules\keycloakify\bin\keycloakify\keycloakify.js:37:53)
    at fulfilled (D:\Temp\keycloakify-starter\node_modules\keycloakify\bin\keycloakify\keycloakify.js:28:58)
```

willwill96:
I think this line may be iterating over the incorrect array:
https://github.com/keycloakify/keycloakify/blob/main/src/bin/tools/trimIndent.ts#L7

PR #286, opened 2023-03-30
fix: pass only strings to trimIndent

willwill96:
Fixes: https://github.com/keycloakify/keycloakify/issues/285


willwill96:
I was going to add some unit tests around the jar functionality, but it would have involved mocking some imports which I think is beyond the capabilities of tsafe. Would you accept a PR that adds the ability to test with jest?

src/bin/tools/jar.ts
```diff
@@ -90,5 +90,5 @@ if (require.main === module) {
             groupId: process.env.GROUP_ID ?? "group",
             version: process.env.VERSION ?? "1.0.0"
         });
-    main().catch(e => console.error(e));
```
willwill96, line 93: Catching this error was causing keycloakify to swallow errors & instead report a successful message & error code:
image

garronej: Yes 💯  you would **never** catch me doing this, I am a vocal advocate for fail early strategy, I just didn't review.

garronej:
Thank you!! 
You guys rocks. I don't have time to handle it myself but you saved me.

garronej:
> I was going to add some unit tests around the jar functionality, but it would have involved mocking some imports which I think is beyond the capabilities of tsafe. Would you accept a PR that adds the ability to test with jest?   

First of [`tsafe`](https://github.com/garronej/tsafe) isn't a test runner, nor a testing framwork. It's a general pupose utility library.  
In particular, it enable to write assertion of type (assert two types are the same, assert one type extends another) which is extremely usefull to write type level unit test like [this one](https://github.com/keycloakify/keycloakify/blob/main/test/lib/tools/AndByDiscriminatingKey.type.ts) (Test that don't have to be run, it it compiles it passes).  

Now for the runtime test it's stupid not having set a testing framwork from the get go an of course I would accept a PR that do that.  
It would avoid having a a [convoluted process to compile everything in CSJ for testing](https://github.com/keycloakify/keycloakify/blob/3bcc6bdf9315ab7f0811119fa6a52128c8d467bf/package.json#L15).  
That said I would prefer if it was [Vitest](https://vitest.dev/), as this is what I'm used to on my [other](https://github.com/garronej/denoify) [projects](https://github.com/codegouvfr/react-dsfr) but no problem if you are more familiar with Jest, I won't be picky!

lordvlad:
<3

src/bin/tools/jar.ts
lordvlad: oh damn, that will affect other placeses too

PR #287, opened 2023-03-30
Vitest integration

willwill96:
- Adds testing supporting via [Vitest](https://vitest.dev/)
- Converts existing tests to be run with Vitest
- Adds input/output build directory options to the keycloakify build options
- Adds a new script `test:keycloakify-starter` which can be used to test the latest version of `keycloakify-starter` with the local changes you have


test/lib/tools/AndByDiscriminatingKey.type.spec.ts
```diff
@@ -0,0 +1,94 @@
+import { AndByDiscriminatingKey } from "keycloakify/tools/AndByDiscriminatingKey";
+import { assert } from "tsafe/assert";
+import type { Equals } from "tsafe";
+import { it, describe } from "vitest";
+
+// These test case names are intentionally vague, because I'm not sure what each are testing individually
+describe("AndByDiscriminatingKey", () => {
+    it("Test Case 1", () => {
```
willwill96, line 8: Should update the test cases in this file

test/bin/jar.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+        };
+    };
+
+    return {
+        // @ts-ignore
+        ...(await vi.importActual("stream")),
+        Readable: readableMock()
+    };
+});
+describe("jar", () => {
```
willwill96, line 22: I thought this test was going to look cleaner, but after mocking imports I'm not sure how valuable it is. I can remove if wanted.

test/lib/tools/AndByDiscriminatingKey.type.spec.ts
garronej: I don't mind the file being renames but I would avoid having type level test files registered to be ran by the test runner.  
This test is not mean to be ran. If it compiles it passes. We just need to make sure this test is tracked by tsconfig.  
I'll handle it!

scripts/test-keycloakify-starter.ts
garronej, line 1: I'm loving it!

garronej:
Hi @willwill96,  
Just a few thing that I noticed on this PR. 

1) We have to be carefull with path, we need to support windows.  
2) Coding style ( JSON style object declaration:  always `"` on keys)  
3) Naming convention `appInputPath` `keycloakBuildPath` are lack consistency, for example we already had established `reactAppBuildDirPath`.  
4) We parse the build option all in one place. the `src/bin/keycloakify/build-path.ts` make a lot of function that where previously pure and easy to reason about dependent to the CWD.  
5) There is almost never a good usecase for `else`  
6) Never use non boolean as boolean ( `path === undefined` instead of `!path`)   
7) In the `bin` directory you can use the `path` module.  
8) When introducing new build options they need to be documented. https://github.com/keycloakify/docs.keycloakify.dev/blob/v7/build-options.md  
9) I'd rather avoid upper case const at the top of the file.  
10) For top level function we use `function ` istead of arrow funct ( because it's the convention of the Deno codebase and function are evaluated first.)  

I refactored:  https://github.com/keycloakify/keycloakify/commit/53e38336fbf6ae81f04cff5e25dc2ae647959eef

willwill96:
Sorry for those mistakes and thank you for extra context @garronej!

A couple points of clarification:
2. Just curious, what is the reasoning for this? If it's just code styling, I wonder if prettier could enforce this: https://prettier.io/docs/en/options.html#quote-props
4. I found it easier to reason about having all the logic relating to build paths in one spot without needing to worry about potential circular dependencies, but that may just be personal preference. I agree it makes more sense to expose the options through `readBuildOptions` though.
8. Ah okay, I was actually wondering where the docs lived. Side note: have you considered converting `keycloakify`, `keycloakify-starter` and the docs into a single mono-repo setup? I understand this might take some special considerations but I think it would make it easier to make changes like this (1 PR vs worst case 3 PRs). I can see value in keeping the starter repo separate to make cloning/forking easier, but I could also see it being handled by a generator (ex yeoman).
9. Is the preference camelcase and when necessary instead?

garronej:
> Sorry for those mistakes and thank you for extra context @garronej!  

I wouldn't call this mistakes, your PR was legit one of the best I ever revied, and I review a lot. Just some little adjustments if we are going to collaborate.  

> 2. Just curious, what is the reasoning for this? If it's just code styling, I wonder if prettier could enforce this: https://prettier.io/docs/en/options.html#quote-props  

Unfortunately there is no prettier option to enforce this, `"preserve"` is as close as it gets.  
I know this coding style is not standard but it makes my life easyer.  
I'm dislexic and without this syntax I have a hard time distinguishing type definition and object instantiation.  
JSON style object declaration make it clearer, it makes keys orange.  

> 4. I found it easier to reason about having all the logic relating to build paths in one spot without needing to worry about potential circular dependencies, but that may just be personal preference. I agree it makes more sense to expose the options through readBuildOptions though.  

I explained my point poorly. 
Look at it this way, for every function we have to be carefull of a few thing:  
- Does the function depend of the `cwd`.  
- Does the function read/write on the fs.  

I would like that the only function that read `process.cwd()` be the `main()`, I would like the `cwd` to be passed as parametter (ad `projectRootPath`) to every function that needs it.  
I pefer having a single function that parse all the options, assigns default ect.  
I don't think there is any circular dependencies in the current state of things.   

> 8. Ah okay, I was actually wondering where the docs lived. Side note: have you considered converting keycloakify, keycloakify-starter and the docs into a single mono-repo setup? I understand this might take some special considerations but I think it would make it easier to make changes like this (1 PR vs worst case 3 PRs). I can see value in keeping the starter repo separate to make cloning/forking easier, but I could also see it being handled by a generator (ex yeoman).  

If the Markdown files was actually where I wrote the doc I would do it but I redact the documentation using GitBook's what you see is what you get editor. It saves a ton of time.  I sent you an invite in DM.  

> 9. Is the preference camelcase and declared when necessary instead?  

Yes, this is just a matter of taste. Also whenever possible I do not create variable when it's not used more than once.  


Thanks for taking the time to give feeback on this.

PR #289, opened 2023-03-30
Some-minor-fixes

lordvlad:
small addendums to previous merge:

1. use path.sep to be cross-platform instead of using OR on two different regex
2. improve typing on trimIndent

TODO

* [x] add unit tests for trimindent


scripts/generate-i18n-messages.ts
```diff
@@ -35,11 +35,10 @@ async function main() {
 
     {
         const baseThemeDirPath = pathJoin(tmpDirPath, "base");
+        const re = new RegExp(`^([^\\${pathSep}]+)\\${pathSep}messages\\${pathSep}messages_([^.]+).properties$`);
```
garronej, line 38: Cool thanks!

PR #290, opened 2023-03-30
refactor: use yauzl for unzipping

mkreuzmayr:


src/bin/tools/unzip.ts
```diff
@@ -1,184 +1,89 @@
...
+            if (err) {
+                reject(err);
+                return;
+            }
 
-function sortByFolderDepth(order: "asc" | "desc") {
-    const ord = order === "asc" ? 1 : -1;
-    return (a: string | undefined, b: string | undefined) => ord * depth(a ?? "") + -ord * depth(b ?? "");
-}
+            zipfile.readEntry();
```
lordvlad, line 39: not certain about the mechanics of yazul, but reading this, I would've assumed `zipfile.readEntry()` should be located _below_ the entry handler

src/bin/tools/unzip.ts
```diff
@@ -1,184 +1,89 @@
...
+import fsp from "node:fs/promises";
+import fs from "fs";
+import path from "node:path";
+import yauzl from "yauzl";
+import stream from "node:stream";
+import { promisify } from "node:util";
+
+const pipeline = promisify(stream.pipeline);
+
+async function pathExists(path: string) {
```
lordvlad, line 10: this is already present in downloadAndUnzip.ts, maybe better  extract?

src/bin/tools/unzip.ts
```diff
@@ -1,184 +1,89 @@
...
-            ...success.map(path => unlink(path).catch(_unused => undefined)),
-            ...failure.map(e => e && e.path && unlink(e.path as string).catch(_unused => undefined))
-        ]);
-        await Promise.all(dirsCreated.filter(Boolean).sort(sortByFolderDepth("desc")));
-        const e = new Error("Failed to extract: " + failure.map(e => e.message).join(";"));
-        (e as any).cause = failure;
-        throw e;
+export async function unzip(file: string, targetFolder: string, unzipSubPath?: string) {
+    // add trailing slash to unzipSubPath and targetFolder
+    if (unzipSubPath && (!unzipSubPath.endsWith("/") || !unzipSubPath.endsWith("\\"))) {
```
lordvlad, line 24:
```suggestion
    if (unzipSubPath && (!unzipSubPath.endsWith(path.sep))) {
```

src/bin/tools/unzip.ts
```diff
@@ -1,184 +1,89 @@
...
+        unzipSubPath += "/";
     }
 
-    return success;
-}
+    if (!targetFolder.endsWith("/") || !targetFolder.endsWith("\\")) {
```
lordvlad, line 28:
```suggestion
    if (!targetFolder.endsWith(path.sep)) {
```

src/bin/tools/unzip.ts
```diff
@@ -1,184 +1,89 @@
...
+        targetFolder += "/";
```
lordvlad, line 29:
```suggestion
        targetFolder += path.sep;
```

src/bin/tools/unzip.ts
```diff
@@ -1,184 +1,89 @@
...
-            ...failure.map(e => e && e.path && unlink(e.path as string).catch(_unused => undefined))
-        ]);
-        await Promise.all(dirsCreated.filter(Boolean).sort(sortByFolderDepth("desc")));
-        const e = new Error("Failed to extract: " + failure.map(e => e.message).join(";"));
-        (e as any).cause = failure;
-        throw e;
+export async function unzip(file: string, targetFolder: string, unzipSubPath?: string) {
+    // add trailing slash to unzipSubPath and targetFolder
+    if (unzipSubPath && (!unzipSubPath.endsWith("/") || !unzipSubPath.endsWith("\\"))) {
+        unzipSubPath += "/";
```
lordvlad, line 25: maybe add a comment that here we must not use path.sep, cause in a zipfile, _only_ forward slashes are valid path separators?

src/bin/tools/unzip.ts
```diff
@@ -1,184 +1,89 @@
...
-                return inflate;
-            }
-            return input;
-        };
-        if (end > start) yield { path: filename, createReadStream: createRecordReadStream, compressionMethod };
-        // advance pointer to next central directory entry
-        i += 46 + filenameLength + extraLength + commentLength;
-    }
+                    // Remove the unzipSubPath from the file name
+                    entry.fileName = entry.fileName.substring(unzipSubPath.length);
```
lordvlad, line 50: not a fan of this mutation

maybe better declare `let target = path.join(targetFolder,entry.filename)` above the if body and correct `target = path.join(targetFolder, entry.filename.substring(unzipSubPath.length))` inside the if body?

or 

put it in an iif? (I've seen @garronej using this thechnique quite extensively so might fit the repo

```
const target: string | undefined = (() => {
  if (unzipSubPath) {
    if (!entry.sfileName.startsWIth(unzipSubPath)) {
      zipfile.readEntry()
      return
    }
    return path.join(targetFolder, entry.filename.substring(unzipSubPath.length)
  }
  return path.join(targetFolder, entry.filename)

**Review by lordvlad — APPROVED** [review]
<3 All comments optional

File: src/bin/tools/unzip.ts (hunk 7 above)
Comment (garronej) [reply src/bin/tools/unzip.ts:50]: I'd write 
```ts
const targetPath: string | undefined = (() => {
  if (unzipSubPath !== undefined) {
    if (!entry.sfileName.startsWIth(unzipSubPath)) {
      return undefined;
    }
    return path.join(targetFolder, entry.filename.substring(unzipSubPath.length)
  }
  return path.join(targetFolder, entry.filename)
})();

if( targetPath === undefined ){
          zipfile.readEntry()
}
```

I avoid side effects in closures.  I never use non boolean as boolean and I aways explicitely `return undefined` otherwise yes.

garronej:
Thanks for that :) :)  

Tests are failing though.  
image

I let you guys pull and see what's up.

willwill96:
I think we may just need to bump the timeouts on those tests. I ran the tests locally successfully, but I did notice that they took a bit longer with this implementation. Running the tests in question on the master version takes ~6-8 seconds(on my machine), whereas with these changes it takes around ~16-20 seconds to run. I'm guessing the CI may just need a bit more time

garronej:
Green acroass the board let's go!  
image

src/bin/tools/unzip.ts
mkreuzmayr: this is not the same, because `path.sep` is specific to the os and not to the string that is passed. We do not know if `/` or `\` paths are passed.

mkreuzmayr:
> While you're at it, maybe you can also refactor zip.ts :D if not ill take swing on the weekend

I am kinda busy right now. `yazl` should be pretty easy to add as we already have a recursive crawler in keycloakify. That functionality could be reused.

src/bin/tools/unzip.ts
mkreuzmayr: 💯

mkreuzmayr: actually I didn´t want to touch anything other than the `unzip.ts`. Adding something like a `fs-utils` folder wouldn´t be a bad idea.

src/bin/tools/unzip.ts
mkreuzmayr: I am not a big fan of this inplace mutation either tbf. Was a bit of mix of "fixing it ASAP" and "yes it works". 😗

src/bin/tools/unzip.ts
lordvlad: good point

PR #296, opened 2023-04-02
Introduce-yazl

lordvlad:
refactor(jar): introduce yazl for creating jars

* introduce yazl
* remove old zip code
* refactor jar code to make it better testable
* introduce unit test for jar creation


test/bin/jar.spec.ts
```diff
@@ -1,26 +1,86 @@
...
+        zipFile.on("end", () => resolve(entries1));
+        zipFile.on("error", e => reject(e));
+        zipFile.readEntry();
+    });
+}
+
 describe("jar", () => {
-    it("creates jar artifacts without error", () => {
-        jar({ artifactId: "artifactId", groupId: "groupId", rootPath: "rootPath", targetPath: "targetPath", version: "1.0.0" });
+    it("creates jar artifacts without error", async () => {
```
lordvlad, line 60: I'm a little on the fence about this test; on the one side, it uses `Buffer` instead of reading from the file system, so it won't cover the use exactly. In my repos, I'd go forth and create a temp dir, not sure what implications that might have for the CI, so I deferred that. @garronej if you agree, I'll add another test which works with actual files instead of testing in-memory

garronej:
Yeah!!!  
TGreatTM  
Thank you!

test/bin/jar.spec.ts
garronej: Of course, you have my approval for that.

garronej:
HI @lordvlad,  I'm getting this error:  

image

Rolling back.

garronej:
Hey @lordvlad, thanks for the fix, there is still an error though:  

image

PR #302, opened 2023-04-05
feat: add keycloakify json schema

willwill96:
# Description
Uses [zod-to-json-schema](https://www.npmjs.com/package/zod-to-json-schema) to add a script which generates a json schema for keycloakify apps to support intellisense. Ex:

image

# Example Usage
```
// package.json
{
    "$schema": "https://raw.githubusercontent.com/keycloakify/keycloakify/4a7dd64982711f388a25ffcb09c8ca8fa91d0f00/keycloakify-json-schema.json",
    "name": "keycloakify-starter"
    ....

```

# Open Questions

~If descriptions for fields are important, how do they get injected?~

Answered: Zod already has [describe](https://zod.dev/?id=describe) which `zod-to-json-schema` supports

### DIFF
--- a/keycloakify-json-schema.json
+++ b/keycloakify-json-schema.json
@@ -1,0 +1,91 @@
+{
+    "allOf": [
+        {
+            "$ref": "https://json.schemastore.org/package.json"
+        },
+        {
+            "$ref": "keycloakifyPackageJsonSchema"
+        }
+    ],
+    "$ref": "#/definitions/keycloakifyPackageJsonSchema",
+    "definitions": {
+        "keycloakifyPackageJsonSchema": {
+            "type": "object",
+            "properties": {
+                "name": {
+                    "type": "string"
+                },
+                "version": {
+                    "type": "string"
+                },
+                "homepage": {
+                    "type": "string"
+                },
+                "keycloakify": {
+                    "type": "object",
+                    "properties": {
+                        "extraPages": {
+                            "type": "array",
+                            "items": {
+                                "type": "string"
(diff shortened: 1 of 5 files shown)


garronej:
@willwill96 Yoooo this is lit!  
We should publish the shema to `https://keycloakify.dev/X.Y.Z/package.shema.json`....  

Or maybe now Keycloakify is mature enough to have its own `denoify.config.{js,json}` like denoify has: https://docs.denoify.land/build-options  

```js
// @ts-check

/** @type { import('keycloakify').BuildOptions } */
const config = {
   //...your options
}

module.exports = config;
```  

We could also submit our logo to https://github.com/vscode-icons/vscode-icons  

I thin that would help establishing the library.  

What do you think?  

The problem whit the JSON schema is that it only works for Keycloakify.  
I think we can reuse the work @PoolOfDeath20 did no Denoify, that would be a quick win.  

What do you think? @willwill96   

Thanks anyway this is real nice, anything that improve typesafety is always welcomed.

willwill96:
@garronej you are absolutely right. I think having a dedicated keycloakify.config file would be cleaner and provide more value. 
If there's some way we could parse a typescript file and inject relevant kcContext info (like extra theme properties), I think that would be even better. I'm not exactly sure how that might work though..

garronej:
> If there's some way we could parse a typescript file and inject relevant kcContext info (like extra theme properties), I think that would be even better. I'm not exactly sure how that might work though..  

Not easy but it would be cool

PR #304, opened 2023-04-06
Fix-broken-jar

lordvlad:
jar was not fixed after introduction of yazl; not  it should; added another proper test with actual files for good measure


garronej:
Thanks again for fixing on such short notice !

lordvlad:
> Thanks again for fixing on such short notice !

I break it, I fix it.

Issue #300, opened 2023-04-04
Typescript error for algToKeyUriAlg in LoginConfigTotp.tsx


thosil:
Hello,

first of all: awesome project!

Typescript is not happy with this line:
https://github.com/keycloakify/keycloakify/blob/a02c38ac4502163935a4c36fcf3efdbb52892a6d/src/login/pages/LoginConfigTotp.tsx#L19

The error: `TS2702: 'KcContext' only refers to a type, but is being used as a namespace here.`

I think it should be:
```ts
const algToKeyUriAlg: Record<typeof kcContext["totp"]["policy"]["algorithm"], string> = { 
```

What do you think?

thosil:
By the way, the page LoginConfigTotop is crashing because it try to call "map()" on totp.policy.supportedApplications, but supportedApplications is directly under totp.

I think it should be defined like this in keycloakify/src/login/kcContext/KcContext.ts
```ts
        totp: {
            enabled: boolean;
            manualUrl: string;
            policy: {
                algorithm: "HmacSHA1" | "HmacSHA256" | "HmacSHA512";
                digits: number;
                lookAheadWindow: number;
            } & (
                | {
                      type: "totp";
                      period: number;
                  }
                | {
                      type: "hotp";
                      initialCounter: number;
                  }
            );
            qrUrl: string;
            supportedApplications: string[];
            totpSecret: string;
            totpSecretEncoded: string;
            totpSecretQrCode: string;
            otpCredentials: { id: string; userLabel: string }[];
        }
```

Here a copy of the totp object when I debug:
```json
{
    "enabled": false,
    "totpSecretEncoded": "...",
    "totpSecret": "...",
    "manualUrl": "https://....",
    "supportedApplications": [
        "totpAppGoogleName",
        "totpAppFreeOTPName"
    ],
    "totpSecretQrCode": "....",
    "policy": {
        "initialCounter": "0",
        "type": "totp",
        "codeReusable": false,
        "lookAheadWindow": "1",
        "algorithm": "HmacSHA1",
        "period": "30",
        "algorithmKey": "SHA1",
        "digits": "6"
    },
    "qrUrl": "https://....",
    "otpCredentials": []
}
```

thosil:
Sure, but I also hit #262, @sukso96100 seems to be working on it.
Should I only fix this one or try to do both in the same PR?

garronej:
If you could do both it would be awesome

thosil:
[...]

PR #307, opened 2023-04-12
fix typing of algToKeyUriAlg + totp interface + cleanup build dir

thosil:
As discussed in #300 there are a couple of fixes in this PR.

1. Infer algToKeyUriAlg from typeof kcConfig  instead of KcConfig, because in a project using keycloakify, KcConfig is not a Namespace anymore ([for good reasons](https://github.com/keycloakify/keycloakify/issues/300#issuecomment-1499325144)).
2. Moved "supportedApplications" to the right place in totp interface.
3. Introduced new build options "keepBuildDir" which is false by default, because the build_keycloak dir is constantly growing. That's because react builds generate new unique assets.

Also:
- add more yarn directories to .gitignore (I'm using yarn 2)
- removed "$" in button name in LoginConfigTotp.tsx

Let me know if something should be fixed.


garronej:
I'm very sorry @thosil, I accidentaly deleted `main` tonight.  
Thank god I had a local copy but it permantently closed this.  
I've re submited the PR here: #325  

Lesson learned: Don't copy paste GPT4 command blindly, however convincing they may be.

PR #337, opened 2023-04-26
Fix/restore missing pom xml in jar

lordvlad:
it was noticed that the `pom.xml` file was missing when using the keycloakify jar bundler. this has been fixed


src/bin/tools/jar.ts
```diff
@@ -65,15 +65,23 @@ export async function jarStream({ groupId, artifactId, version, asyncPathGenerat
  * Create a jar archive, using the resources found at `rootPath` (a directory) and write the
  * archive to `targetPath` (a file). Use `groupId`, `artifactId` and `version` to define
  * the contents of the pom.properties file which is going to be added to the archive.
+ * The root directory is expectedto have a conventional maven/gradle folder structure with a
```
lordvlad, line 68:
```suggestion
 * The root directory is expected to have a conventional maven/gradle folder structure with a
```

garronej:
Many thanks!

Issue #336, opened 2023-04-26
Issue with the JARs created by the Keycloakify bundler


dyeske61283:
Hello,

As pointed out here "https://docs.keycloakify.dev/build-options#keycloakify.bundler" we should create an issue, if we need to use the legacy **mvn** bundler. This was the case.

**Our use case & issue:**
When using the keycloakify build, we need to unpack the generated JARs in the process to add to some additional files and perform some renaming. After using the keycloakify bundler, the JAR files could not be unpacked by using either JAR or unzip -- the errors are added at the bottom. 
This is apparently due to the format of the created archive. See the thread here for an explanation: https://stackoverflow.com/questions/47208272/android-zipinputstream-only-deflated-entries-can-have-ext-descriptor 

Workaround:
Using **mvn** as the bundler works fine.

---
This was the error, when using **JAR**:
java.util.zip.ZipException: only DEFLATED entries can have EXT descriptor
	at java.base/java.util.zip.ZipInputStream.readLOC(ZipInputStream.java:313)
	at java.base/java.util.zip.ZipInputStream.getNextEntry(ZipInputStream.java:125)
	at jdk.jartool/sun.tools.jar.Main.extract(Main.java:1360)
	at jdk.jartool/sun.tools.jar.Main.run(Main.java:409)
	at jdk.jartool/sun.tools.jar.Main.main(Main.java:1680)

This was the error, when using **unzip**: 
_error: invalid zip file with overlapped components (possible zip bomb)_

lordvlad:
@dyeske61283 anyway tofind out which jar entry caused the error? Can't reproduce the unzip issue. Can you try unzip again with the zip bomb detection disabled? `UNZIP_DISABLE_ZIPBOMB_DETECTION=true unzip -t bla.jar`

lordvlad:
Quite certain this is related to #330 #331

dyeske61283:
@lordvlad The unzip did work for me as well on my local machine (MacOS), but not inside a node:18.12.1 docker image (which should be debian-based in our build pipeline). I will disable the detection and try it again.

In regards to the JAR entries: Sadly, starting the JAR executable in verbose mode does not reveal any additional info.

lordvlad:
@dyeske61283 can you try using the branch from #339?

garronej:
@dyeske61283 I just released, you can test with latest. No need to checkout the branch.  
[...]

Issue #331, opened 2023-04-23
Mime Type Error on latest keycloakify-starter


willwill96:
My login pages aren't rendering and getting this on the latest keycloakify-starter repo:

![image](https://user-images.githubusercontent.com/10997562/233870153-860b5c7a-b5cc-4c74-9114-a984cdfdb26c.png)

This is a fresh install of keycloak + keycloakify. Here is my Dockerfile for reference (Located one directory up from keycloakify-starter):
```Dockerfile
FROM node:18 as keycloakify_jar_builder

COPY ./keycloakify-starter/package.json ./keycloakify-starter/yarn.lock /opt/app/

WORKDIR /opt/app

RUN yarn install --frozen-lockfile

COPY ./keycloakify-starter/ /opt/app/

RUN yarn build-keycloak-theme

FROM quay.io/keycloak/keycloak:latest as builder

WORKDIR /opt/keycloak

COPY --from=keycloakify_jar_builder /opt/app/build_keycloak/target/keycloakify-starter-keycloak-theme-4.4.4.jar /opt/keycloak/providers/
RUN /opt/keycloak/bin/kc.sh build


FROM quay.io/keycloak/keycloak:latest
COPY --from=builder /opt/keycloak/ /opt/keycloak/
ENV KC_HOSTNAME=localhost
ENTRYPOINT ["/opt/keycloak/bin/kc.sh", "start-dev"]
```

I verified stale assets aren't being included and I've tried with both building the keycloakify jar inside and out of a Dockerfile build step

garronej:
@willwill96  😥   

I can't reproduce, I just tried right now, everything is working as expected on my end (except https://www.keycloak.org/app is down).  

I'm going to paste here the same aswer I gave you yesterday, I'm very willing to find what's up but I'll need a clear reproduction path.  

Hello @willwill96!

That's odd indeed.

Can you confirm if you built your theme with the `--external-assets` option (please check your CI workflow)? I don't believe you did, but if so, that could easily explain the error you're encountering.

I'm currently using the latest version in production, and it works perfectly:

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/6702424/233735475-468e3c4f-ea98-4c39-9acc-722189cf5813.png">

Additionally, the latest Keycloakify starter theme is also running smoothly in production:

<img width="1724" alt="image" src="https://user-images.githubusercontent.com/6702424/233735584-58c86700-cd35-4bf1-8223-e10f88893e97.png">

[...]

PR #339, opened 2023-04-28
fix: fix broken jar

### LINKED ISSUE #330: Failed to load main.hash.js
We just tried updating to the latest version and got the following error.

Screenshot 2023-04-21 at 2 05 49 PM

Our last known working version is `7.6.3`

lordvlad:
Many tools will handle zipfiles which lack directory entries just fine, others will not. Looks like the JDKs JAR libs are not handling  it well. This commit will make sure to create folder entries.

Should fix #330 and #331; Probably also #336, though I'd need some confirmation there since the reproduction path seems rather elaborate


garronej:
Fantastic!  
Thank you so much for investigating this!

PR #348, opened 2023-05-15
Parse datetime objects as iso strings

kpoelhekke:
This parses datetime objects as iso strings.

This is needed for example on the `sessions.ftl` page where you want to render the `started`, `lastAccess` and `expires` values.


garronej:
@all-contributors please add @kpoelhekke for code

garronej:
Relasing new version

PR #350, opened 2023-05-23
Change node.js 16.6.0 dependency that Array.prototype.at

abdurrahmanekr:
My node.js environment is v14.19.1. Because of that i get this error:

```
TypeError: strings[i].split(...).at is not a function
at populateTemplate (/path/foo/keycloakify-starter/node_modules/keycloakify/bin/tools/trimIndent.js:46:75)
```

### DIFF
--- a/src/bin/tools/trimIndent.ts
+++ b/src/bin/tools/trimIndent.ts
@@ -9,7 +9,7 @@ function populateTemplate(strings: TemplateStringsArray, ...args: unknown[]) {
         if (strings[i]) {
             chunks.push(strings[i]);
             // remember last indent of the string portion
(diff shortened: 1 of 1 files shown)


abdurrahmanekr:
I understand you @lordvlad . As you know some environments needs very big change. I wish have a chance for changing all environments. Can typescript convert javascript to 14? Maybe such a setting would solve the whole problem, in this way we don't need to use old javascript methods

lordvlad:
https://www.typescriptlang.org/tsconfig#target

garronej:
@abdurrahmanekr I have no problem with that. I'm open to anything that can make your life easier.

Thank you for your contribution. I'm releasing a new version. Please feel free to reach out if you encounter any other issues or have any further problems of this nature.

Issue #369, opened 2023-06-27
Error on messagesPerField evaluation in error.ftl


abencun-symphony:
First of all thank you for your active maintenance of this awesome project!

The following exception occurs in `error.ftl`:

```
The following has evaluated to null or missing:
==> messagesPerField  [in template "error.ftl" at line 42, column 27]

----
Tip: If the failing expression is known to legally refer to something that's sometimes null or missing, either specify a default value like myOptionalVar!myDefault, or use <#if myOptionalVar??>when-present<#else>when-missing</#if>. (These only cover the last step of the expression; to cover the whole expression, use parenthesis: (myOptionalVar.foo)!myDefault, (myOptionalVar.foo)??
----

----
FTL stack trace ("~" means nesting-related):
	- Failed at: #if !messagesPerField.existsError??  [in template "error.ftl" at line 42, column 21]
```

We're not really sure why or how is this happening - but looking at this error message it looks to us that perhaps `!messagesPerField.existsError??` should be changed to `!(messagesPerField.existsError)??` or a similar expression?

Also, I see that this code seems to be present in other ftl templates and runs fine - maybe it's something configuration-related on our end that we should be paying attention to?

We're running Keycloakify 7.12.8 on Keycloak v20.0.5.

garronej:
Hello @abencun-symphony,  

Thank you for reporting, be sure I'll adress the issue in good delay.  
The exception is only loged by Keycloak but otherwise, the page still works correct?  
Can you try with the latest Keycloakify version?  
If you still have the error, can you try with [7.11.10](https://github.com/keycloakify/keycloakify/releases/tag/v7.11.10)?  

Thanks in advance

abencun-symphony:
I think that the page runs correctly. Will come back with more details. Thank you for a quick response!

abencun-symphony:
@garronej 7.11.10 and latest don't fix the issue. However I think I've tracked it down and I've opened PR #370 with a potential fix.

garronej:
Ohhhh, I see, I did a big mistake.  
I got confuse between the JavaScript contrôle flow and the FreeMarker contrôle flow.  
Writing FTL code that output JavaScript is chalenging to reason about.  

Anyway, I'm releasing 7.14.2 that should fix the issue. Thanks for reporting!

abencun-symphony:
Awesome, appreciate it!

garronej:
This must have been fixed long ago

PR #370, opened 2023-06-27
Conditionally use messagesPerField.

abencun-symphony:
It seems like on pages like `error.ftl` the `messagesPerField` variable is not present and therefore FTL fails to render properly. This PR wraps all of the usages of messagesPerField in an `else` block. I've described this behavior in #369 and this PR tries to remediate this issue.

I'm absolutely unskilled in writing FTL so if anyone has a better solution to this issue or perhaps the core of the issue lies elsewhere and could be fixed in a better way please do let me know. Furthermore, I'm not sure of whether this change could break something - for us it didn't break any pages but someone should really take a deeper look at these changes.


abencun-symphony:
Now that I'm looking at this code maybe wrapping the code in an `<#else>` block would've been enough since there's already a check whether `messagesPerField` is present? Waiting on more input here

Issue #394, opened 2023-08-10
Do not mock message in kcContextCommonMock


ddubrava:
Hi! Is it possible to avoid [mocking the message in kcContextCommonMock](https://github.com/keycloakify/keycloakify/blob/5156b2e0cc6b2834b19837a8888c3c27c4b0415a/src/login/kcContext/kcContextMocks.ts#L237)? The issue is that we have to turn it off for each page or pass `{ message: undefined }` for every story. Both options are inconvenient, since we end up disabling it in 99% of cases

garronej:
Sure!  
Do you want to submit a PR for it?  
Otherwise I'll do it.

ddubrava:
> Sure! Do you want to submit a PR for it? Otherwise I'll do it.

yeah, I'll do it. By the way, why is it enabled? I mean, what's the idea behind it?

garronej:
Great thanks!

It is enabled by default because only recently did we introduce Storybook with various scenarios for each page. Before that, you would just uncomment mockPageId and test by running yarn start. In this scenario, I thought that if I didn't put this message by default, no one would know this how to make it appear.

PR #396, opened 2023-08-11
feat: remove message from kcContextCommonMock

ddubrava:
Closes https://github.com/keycloakify/keycloakify/issues/394


ddubrava:
I think we need also to update the Storybook starter

PR #399, opened 2023-08-14
fix: usernameEditDisabled renamed to usernameHidden

### LINKED ISSUE #397: KcContext: usernameEditDisabled has been renamed to usernameHidden
Hello, first off, thanks a lot for your awesome work !

I'd like to report the following issue :
The `usernameEditDisabled` property has been renamed to `usernameHidden` a few months ago and the general display conditions for the username field have changed as a result in the base `login.ftl` template.
Source : https://github.com/keycloak/keycloak/commit/d9c8cb30a5ec2f4a9589cdbfe50bf8c35d9853ad#diff-211584f3e252b06de72bea35ec29af7e7c7236d4094d25eb121f62c1cd41b4f8

Could you please look into it ?

DevRebs:
- update KcContext definition for login.ftl
- update the username field rendering conditions in Login.tsx

Closes #397

### DIFF
--- a/src/login/kcContext/KcContext.ts
+++ b/src/login/kcContext/KcContext.ts
@@ -145,7 +145,7 @@ export declare namespace KcContext {
             rememberMe?: string;
             password?: string;
         };
-        usernameEditDisabled: boolean;
+        usernameHidden?: boolean;
         social: {
             displayInfo: boolean;
             providers?: {
--- a/src/login/kcContext/kcContextMocks.ts
+++ b/src/login/kcContext/kcContextMocks.ts
@@ -264,7 +264,7 @@ export const kcContextMocks = [
         "social": {
             "displayInfo": true
         },
-        "usernameEditDisabled": false,
+        "usernameHidden": false,
         "login": {},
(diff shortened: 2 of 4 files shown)


DevRebs:
Not sure what should be done about this story : [immutable preset username](https://storybook.keycloakify.dev/?path=/story/login-login-ftl--with-immutable-preset-username)

garronej:
Thanks!  

So it's no longer possible to have the email presset and non editable?

DevRebs:
The scenario is still valid as it is used / triggered in the case of forced re-authentication when using [`prompt=login` or `max-age=0`](https://www.keycloak.org/docs/latest/securing_apps/#_params_forwarding) (and always was - better explained here https://github.com/keycloak/keycloak/pull/9580).

Use of `login_hint` still makes it possible to pre-fill the `username` field but it was never possible to set it on disabled out of the box through configuration (as far as I can tell).

I have figured out the proper setup for the storybook and updated my branch  :

image

garronej:
Fantastic thank you very much!

PR #407, opened 2023-08-24
Issue 406

garronej:


garronej:
This has to be merged! 
@celinepelletier also reported some errors in the logs that need to be muted!

PR #423, opened 2023-09-27
fix: change JS path transformation for static resources

zavoloklom:
- Handle both arrow functions and traditional function expressions
- Add tests to ensure correctness of transformations
- Remove redundant condition `${true ? "/build/" : ""}`
- Add `.replace(/\s+/g, " ").trim()` to make new code look like minified


garronej:
Hello @zavoloklom,

Thank you for your valuable pull request! Everything looks good to me.

I do find it a bit unconventional that Keycloakify's input is the output of webpack, making it tightly coupled with webpack's implementation. However, the alternative—integrating a bundler within Keycloakify—would present its own set of challenges. Overall, I believe this approach is sustainable thanks to contributions from individuals like you who take the time to incrementally improve the regular expressions. So, a big thank you for your efforts!

Best regards,

garronej:
@all-contributors please add @zavoloklom for tests and code

PR #442, opened 2023-10-26
v6: Use message string for the supported applications

anankul:
Apparently, the totp.policy.supportedApplications array returns the message keys for the apps, they need to be replaced by actual translation. To keep the default behavior before this PR, if the translation is not available, it will default to the message key on screen.

### DIFF
--- a/src/lib/pages/LoginConfigTotp.tsx
+++ b/src/lib/pages/LoginConfigTotp.tsx
@@ -29,7 +29,7 @@ export default function LoginConfigTotp(props: PageProps<Extract<KcContextBase, 
 
                             <ul id="kc-totp-supported-apps">
                                 {totp.policy.supportedApplications.map(app => (
-                                    <li>{ msg(app as MessageKeyBase) == undefined? app : msg(app as MessageKeyBase) }</li>
+                                    <li>{msg(app as MessageKeyBase)}</li>
                                 ))}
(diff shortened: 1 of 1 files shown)


src/lib/pages/LoginConfigTotp.tsx
```diff
@@ -29,7 +29,7 @@ export default function LoginConfigTotp(props: PageProps<Extract<KcContextBase,
 
                             <ul id="kc-totp-supported-apps">
                                 {totp.policy.supportedApplications.map(app => (
-                                    <li>{app}</li>
+                                    <li>{ msg(app as MessageKeyBase) == undefined? app : msg(app as MessageKeyBase) }</li>
```
garronej, line 32:
```suggestion
                                    <li>{msg(app as MessageKeyBase)}</li>
```
No need, it will fallback automatically

garronej: Oh damn very sorry about this, I think it was fixed latter on... I'll adress this hold on.

garronej: Should be fixed on the latest v6 that is beeing published. Sorry for the messy review

anankul: No worries, took the latest, all good!

PR #461, opened 2023-11-30
Fixes #459: Use Template from props

### LINKED ISSUE #459: Pages using Template from import instead of PageProps
The following pages are importing "Template" using a relative import instead of using Template from PageProps:

- src/login/pages/LoginDeviceVerifyUserCode.tsx
- src/login/pages/LoginOauthGrant.tsx

This means that these pages needs to be copied to any project which customizes the Template, even though no other modifications are done on the pages.

BlackVoid:
Use Template instead of importing Template


garronej:
@all-contributors please add @BlackVoid for code

PR #462, opened 2023-11-30
Fixes #460: Fixes KcContext to contain attributes for client object

### LINKED ISSUE #460: KcContext for client missing attributes
The client object in KcContext is missing typing for the attributes field:
```
attributes?: Record<string, string>;
```

BlackVoid:


garronej:
Yes,  
I think we could even loose the `?` am I right?

BlackVoid:
I can remove it, was just unsure if it's always there or not :)

garronej:
Yes it is.  
Thank you for adding this. It's indeed important to show pepole that this exist since it holds the environement variables which are very important when you build a customizable theme.  

Even myself was doing this:  

image

garronej:
Releasing a new version now

PR #475, opened 2023-12-15
Adds recaptcha script management

### LINKED ISSUE #470: Recaptcha doesn't work
Recaptcha doesn't work with the custom Keycloakify theme I made, despite the property "recaptchaRequired" being true & "recaptchaSiteKey" having the appropriate value (using Google Recaptcha). The html element itself appears in the DOM, but for some reason it's not "populated" as expected.

Screenshot 2023-12-12 001121

If for the same client, I swap the Keycloakify theme to a keycloak built-in one, it works completely fine. I have used the starter project (latest Keycloakify version) to build a Keycloak standalone theme, not an app, thus not sure if that has anything to do with it, although it shouldn't pose any problem.

I'd really appreciate any pointer/idea on how to solve this/why this might be happening. TIA.

D-Simona-G:
Fixes #470

### DIFF
--- a/src/login/pages/RegisterUserProfile.tsx
+++ b/src/login/pages/RegisterUserProfile.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import { clsx } from "keycloakify/tools/clsx";
 import { UserProfileFormFields } from "./shared/UserProfileFormFields";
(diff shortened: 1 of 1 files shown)


garronej:
Thanks a lot for your help on this @D-Simona-G,  
Even if I didn't ended up merging because v10 is too big of a refactor, your insights and PR helped a ton!  

image

PR #476, opened 2023-12-18
Fixes #473: Set document title using the "loginTitle" translation

### LINKED ISSUE #473: Missing page title
The title is always "React App" and is never set to anything else, perhaps it would be good to have it in line with the default page titles?

Maybe implement it with something like [this](https://stackoverflow.com/questions/46160461/how-do-you-set-the-document-title-in-react)?
```
import React, { useEffect } from 'react';

function Example() {
  useEffect(() => {
    document.title = 'My Page Title';
  }, []);
}
```

BlackVoid:
Since realm can't be changed without the page being refreshed I think this is a suitable solution.

The starter kit projects needs to adapt this change for people who is using that to benefit from this change.

### DIFF
--- a/src/login/Template.tsx
+++ b/src/login/Template.tsx
@@ -45,6 +45,8 @@ export default function Template(props: TemplateProps<KcContext, I18n>) {
         return null;
     }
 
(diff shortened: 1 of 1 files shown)


garronej:
Great! Thanks!

PR #487, opened 2024-01-12
WebAuthn Register

D-Simona-G:
Adds webauthn-register.ftl page


D-Simona-G:
Oops, I just realised I branched off of the other branch I made a PR previously instead of main.
Will fix it and recreate

PR #489, opened 2024-01-15
refactor: point the import statement to keycloakify instead of internal

law108000:
align import statement with other page templates


garronej:
Yes perfect thank you!  

@all-contributors please add @law108000 for code

PR #502, opened 2024-02-02
fix: radio inputs on login-otp page

giorgoslytos:
Adds support for the case of more than 1 devices registered on the totp.

Previously the inputs were hidden, their name prop was incorrect and there was a bug in the interpolation of their value prop resulting in failure when submitting the form.

Javascript was pruned as it was found to be unnecessary under the new implementation


garronej:
Looks very good to me!  
If it passes, I'll merge it!

garronej:
@all-contributors please add @giorgoslytos for code

garronej:
I just released a new version with your change. Thank you for your contribution @giorgoslytos

PR #515, opened 2024-03-02
feat: addition of extra account pages

giorgoslytos:
Addition of the following account pages:
- sessions.ftl
- totp.ftl
- applications.ftl
- log.ftl


garronej:
I am so hyped!

PR #575, opened 2024-06-28
Added & Fixed Applications Page Under Account

liamlows:
- Added the applications.ftl story page.
- Fixed double comma that would appear if both `realmRolesAvailable` and `resourceRolesAvailable` attributes were present. Simply removed a check that was redundant (additional check added comma on line 69 in original file)


garronej:
Thanks a lot that perfect!

garronej:
@all-contributors please add @liamlows for code and doc

PR #589, opened 2024-07-17
fix: Typo InputFiledByType to InputFieldByType

lokmeinmatz:
Noticed this typo during migrating our Keycloakify Frontend ^^
Thanks for the great work!


garronej:
Thank you very much for this!

garronej:
Note that I also forget a `import { assert } from "tsafe/assert"` you can remplace this line by `import { assert } from "keycloakify/tools/assert";` if you don't want to add tsafe in your dependency list. (I just fixed it)

PR #615, opened 2024-08-21
Fix array comparison and improve type check

oliviergoulet5:
I identified a bug where a comparison would always be false because arrays would be compared by reference.

I also changed the type-checking for an array to use the more semantic `Array.isArray` instead of comparing constructors.


garronej:
Thank you for the PR!  

This is crazy because it's [from Keycloak 24 official release](https://github.com/keycloak/keycloak/blob/6a5783e9d538a1e67512fe43abe7b15ca6eed5c2/themes/src/main/resources/theme/base/login/webauthn-register.ftl#L150-L175) and it's obviously broken.  

This is why it made it in Keycloakify

garronej:
@all-contributors please add @oliviergoulet5 for code

garronej:
I've released a new version

PR #646, opened 2024-09-11
Allow docker start script to work with podman

mkreu:
Hey there,

I noticed that the start-keycloak script refuses to run when the output of `docker --version` does not contain the word "Docker", which is the case when using Podman as a drop-in replacement for Docker.

~~This PR changes the script to only check if the command `docker --version` completes successfully instead of relying on a certain string in the output.~~

This PR changes the script to allow "podman" in the output of `docker --version` as well

### DIFF
--- a/src/bin/start-keycloak/start-keycloak.ts
+++ b/src/bin/start-keycloak/start-keycloak.ts
@@ -37,15 +37,17 @@ export type CliCommandOptions = CliCommandOptions_common & {
 
 export async function command(params: { cliCommandOptions: CliCommandOptions }) {
     exit_if_docker_not_installed: {
-        let commandOutput: Buffer | undefined = undefined;
+        let commandOutput: string | undefined = undefined;
 
         try {
-            commandOutput = child_process.execSync("docker --version", {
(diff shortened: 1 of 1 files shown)


garronej:
Hello thank you for raising this. This is indeed a shame not to support podman.  

However I would like to check that Docker is actually docker.  

Just making sure that the `docker --version` command does not crash isn't enough for me.  

Beleve it or not pepole would just `npm install -g docker` and get [This package](https://www.npmjs.com/package/docker).  



Can you give me the output you're getting when running `docker --version`

mkreu:
> Just making sure that the `docker --version` command does not crash isn't enough for me.
> 
> Beleve it or not pepole would just `npm install -g docker` and get [This package](https://www.npmjs.com/package/docker).

I do see that point.

> Can you give me the output you're getting when running `docker --version`

```sh
$ docker --version
podman version 5.2.2
```

So I guess we could just check for either `Docker` or `Podman` in the string?

garronej:
>  Docker or Podman in the string?  

Yes!

garronej:
Yes! Great thank you!

PR #678, opened 2024-10-02
respect inputOptionLabels

### LINKED ISSUE #677: inputOptionLabels not always respected
When using a `multiselect-checkboxes` or `select-radiobuttons`. i18n and `inputOptionLabels` are ignored.

This works on default keycloak theme.

johanjk:
fixes #677

### DIFF
--- a/src/login/UserProfileFormFields.tsx
+++ b/src/login/UserProfileFormFields.tsx
@@ -434,9 +434,7 @@ function AddRemoveButtonsMultiValuedAttribute(props: {
 }
 
 function InputTagSelects(props: InputFieldByTypeProps) {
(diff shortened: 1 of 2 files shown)


johanjk:
Works within the new story I added, validated it did not work in that story before the changes.

Have exported to full Keycloak with `yarn link-in-starter` then `yarn build-keycloak-theme` and added matching configs for  both `multiselect-checkboxes` and `select-radiobuttons` with `inputOptionLabels` and found it working. 

Have not explicitly check i18n, just assumed that it should work, as we're just doing the same as what is already working for `select`.  

If you want any explicit tests, please guide me to how you would like a test be formulated.

PR #688, opened 2024-10-08
fix: added parameter type for story context on register page

liamlows:
simply added in the missing parameter type that was throwing build errors on the register.stories.tsx file.


garronej:
Thank you!

PR #647, opened 2024-09-11
:sparkles: Added `generate-resources` CLI command

pushrbx:
This is a feature request I believe.
I have a bigger repo of extensions for keycloak, but written in kotlin and instead of maven, gradle is the build system. I'd like to add themes to my repo, and keycloakify seems to be a good fit. However it would feel strange to use maven next to gradle through keycloakify in said repo.
I've figured that if I just make the `keycloakify` cli to just generate the resources, I can create my own gradle script to pick those up and build a `jar` file, instead of relying on `keycloakify` cli to do that for me through maven.

Just for reference my gradle script looks like this:

```kotlin
import java.nio.file.Files
import kotlin.io.path.Path
import org.siouan.frontendgradleplugin.infrastructure.gradle.InstallFrontendTask

plugins {
    kotlin("jvm") version "2.0.20"
    id("com.gradleup.shadow") version "8.3.0"
    id("org.siouan.frontend-jdk21") version "8.1.0"
    id("java")
}

group = "org.pushrbx"
version = "1.0-SNAPSHOT"
val jarBaseName = "org.pushrbx.keycloak-themes"

repositories {
    mavenCentral()
}

dependencies {
    testImplementation(kotlin("test"))
    implementation("org.keycloak", "keycloak-client-registration-api", "25.0.4")
    implementation("dev.jcputney", "keycloak-theme-additional-info-extension", "1.1.5")
    implementation("io.phasetwo.keycloak", "keycloak-account-v1", "0.6")
}

tasks.test {
    useJUnitPlatform()
}

kotlin {
    jvmToolchain(17)
}

java {
    toolchain {
        languageVersion.set(JavaLanguageVersion.of(17))
    }
}

tasks.withType<JavaCompile> {
    options.release.set(17) // Ensure Java compatibility with 17
}

frontend {
    nodeVersion.set("22.6.0")
    nodeInstallDirectory.set(project.layout.projectDirectory.dir(".node"))
    installScript.set("install")
    assembleScript.set("run build")
    cleanScript.set("run clean")
}

tasks {
    val themeFiles by named<InstallFrontendTask>("installFrontend") {
        val ciPlatformPresent = providers.environmentVariable("CI").isPresent()
        val lockFilePath = "${projectDir}/package-lock.json"
        val retainedMetadataFileNames: Set<String>
        if (ciPlatformPresent) {
            retainedMetadataFileNames = setOf(lockFilePath)
        } else {
[...]

Issue #673, opened 2024-09-30
CLI option to skip rmSync resourcesDirPath


jackie-linz:
We have custom keycloak image that's built with our own extensions, and have a docker-compose file for local development environment.

When developing with vanilla Keycloak theme, we could mount the custom themes folder into the running container like the following and any theme change could be reflected without restart.

```yaml
services:
  auth:
    image: "my-keycloak-image"
    command:
      # these parameters to have theme change show up without restarting container
      - "--spi-theme-static-max-age=-1"
      - "--spi-theme-cache-themes=false"
      - "--spi-theme-cache-templates=false"
    volumes:
      - ./my-themes:/opt/keycloak/themes
```

When using keycloakify, it builds a JAR file and immediately deletes the `resourcesDirPath`:

https://github.com/keycloakify/keycloakify/blob/main/src/bin/keycloakify/keycloakify.ts#L128

Please provide a CLI option to skip the `rmSync` call so we can mount the theme folder into the container similar to what we used to do

garronej:
Hello @jackie-linz,

Feel free to use `jar -xf` to extract the jar and mount the resources in your container, but **this is not recommended**.

Extensions should be loaded as jars, and you can use the following command to install the extension into Keycloak:

```bash
bash /opt/keycloak/bin/kc.sh build
```

For more details, refer to the [documentation](https://docs.keycloakify.dev/importing-your-theme-in-keycloak).

Mounting the resources alone won’t work completely (at least not until Keycloak 26 is released). The Keycloakify theme can include polyfills for older Keycloak versions and features that haven’t been released yet (like [this one](https://github.com/keycloak/keycloak/pull/32625)), depending on your [target Keycloak version](https://docs.keycloakify.dev/targeting-specific-keycloak-versions). These polyfills are Java extensions.

Did you know that Keycloakify provides a command to test your theme in a Keycloak container?  
Check it out here: https://docs.keycloakify.dev/testing-your-theme/in-a-keycloak-docker-container

Best regards,

jackie-linz:
Hi @garronej ,

Thanks for the quick reply. Just to clarify, we currently only use keycloakify for the account theme (single page) so the custom extensions that came with keycloakify should hopfully not matter.

We follow this [Keycloak documentation](https://www.keycloak.org/docs/latest/server_development/#creating-a-theme) for login and email theme development and it gives us instant feedback for UI changes.

[...]

Issue #660, opened 2024-09-18
Migration v9 -> v10 - pom.xml, docreatejar


500E:
Hello! 
I want to migrate from v9 to v10 and most of the process is going well. However I'm missing two things that I can't find it the docs: 
1. https://docs.keycloakify.dev/v/v9/build-options#docreatejar - is there an equivalent for this option in v10?
2. `pom.xml` file creation - in v9 it was placed directly in the `dist_keycloak` folder after the build, now I can only find it in the generated `.jar` file. Is it possible to generate `pom.xml` file similarly to the v9?

Thank you!

thomashexagon:
I was also looking for 1.

garronej:
Hello @500E,  

- 1. No, it's no longer possible. The theme dosen't work anymore just by copying the assets manually into Keycloak. The proper way to load extentions in Keycloak is to put a jar in `/opt/keycloak/providers` and run `bash /opt/keycloak/bin/kc.sh build` then start Keycloak.  
- 2. No it's not possible.  

Thoses two things can be addressed just by running `jar -xf keycloak-theme.jar`.  
Maybe you can describe to me what you are trying to acheive and I can tell you what's the best way to do it.

iliascholl-up42:
Hey @garronej,

I am a colleague of @500E. We are injecting `repositories` and `distributionManagement` sections into the `pom.xml` to reference our private maven repository and run `mvn install` and `mvn deploy` to build and publish the JAR to our private repo. The versioned publishing enables us to pull the JAR during the deployment of the Keycloak server.

We are in the process of changing our deployment process and I think we will be able to work around this issue with a monorepo setup.

On a different note: could you please elaborate on the following:
> The theme doesn't work anymore just by copying the assets manually into Keycloak.

garronej:
Hello @iliascholl-up42, 

Can't you use [this](https://github.com/keycloakify/keycloakify/issues/623#issuecomment-2314675038) to publish the JAR after the JAR is built?  

> On a different note: could you please elaborate on the following:

This extension is needed for the password policies to be available: https://github.com/jcputney/keycloak-theme-additional-info-extension  
We got this PR merged in Keycloak so it won't be required anymore starting with Keycloak 26: https://github.com/keycloak/keycloak/pull/32625  
We also need https://github.com/p2-inc/keycloak-account-v1 for pepole implementing [the multi page flavor of the account theme](https://docs.keycloakify.dev/account-theme/multi-page)

[...]

PR #689, opened 2024-10-09
Allow skipping rm Keycloakify dev resource

chenhunghan:
Similar to https://github.com/keycloakify/keycloakify/pull/647 but taking a **_shortcut_** to allow skipping removal of Keycloakify's dev resources. The reason for this is because we need to the raw dev resource to update the static files (js,css) into Keycloak, instead of packing into a `.jar.` and then upload.

### DIFF
--- a/src/bin/shared/buildContext.ts
+++ b/src/bin/shared/buildContext.ts
@@ -93,6 +93,7 @@ export type BuildOptions = {
         realmJsonFilePath?: string;
         port?: number;
     };
+    skipRemoveDevResources?: boolean;
 } & BuildOptions.AccountThemeImplAndKeycloakVersionTargets;
 
 export namespace BuildOptions {
--- a/src/vite-plugin/vite-plugin.ts
+++ b/src/vite-plugin/vite-plugin.ts
@@ -199,16 +199,18 @@ export function keycloakify(params: keycloakify.Params) {
 
             assert(buildDirPath !== undefined);
 
-            await rm(
-                pathJoin(
-                    buildDirPath,
-                    WELL_KNOWN_DIRECTORY_BASE_NAME.KEYCLOAKIFY_DEV_RESOURCES
-                ),
-                {
-                    recursive: true,
-                    force: true
-                }
-            );
+            if (!buildOptions?.skipRemoveDevResources) {
+                await rm(
+                    pathJoin(
+                        buildDirPath,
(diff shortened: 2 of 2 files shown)


garronej:
Hello @chenhunghan,  

I think there is a missunderstanding.  
The repository `public/keycloak-dev-resouce` is only there so you can test your login pages into storybook.  
Nothing more. It should never be included in the build.  

> The reason for this is because we need to the raw dev resource to update the static files (js,css) into Keycloak, instead of packing into a .jar.  

This is not the way to load extention into keycloak.  
https://github.com/keycloakify/keycloakify/issues/673  
https://github.com/keycloakify/keycloakify/issues/660#issuecomment-2360748968  

If you are not implementing a Multi-Page account theme and are deploying on Keycloak 26, copyting the files manually will work... today, but it the future I might need to add java extention so that you theme remain compatible with future version of Keycloak.  
It's best to stick with the official way of loading extention in Keycloak. Just trying to prevent you from shooting yourself in the foot here.

chenhunghan:
> Hello @chenhunghan,
> 
> I think there is a missunderstanding. The `public/keycloak-dev-resouce` directory is only there so you can test your login pages in storybook. Nothing more. It should never be included in the build.
> 
> > The reason for this is because we need to the raw dev resource to update the static files (js,css) into Keycloak, instead of packing into a .jar.
> 
> This is not the way to load extention into keycloak. [#673 (comment)](https://github.com/keycloakify/keycloakify/issues/673#issuecomment-2381914267) [#660 (comment)](https://github.com/keycloakify/keycloakify/issues/660#issuecomment-2360748968)
> 

So this only works in Keycloak > 26?

https://www.keycloak.org/docs/latest/server_development/#deploying-themes

> Themes can be deployed to Keycloak by copying the theme directory to themes or it can be deployed as an archive. During development you can copy the theme to the themes directory, but in production you may want to consider using an archive...

Thanks for the pointer, will use `unzip keycloak-theme-for-kc-22-and-above.jar -d extracted` intead.

> If you are not implementing a Multi-Page account theme and are deploying on Keycloak 26, copyting the files manually will work... today, but it the future I might need to add java extention so that you theme remain compatible with future version of Keycloak. It's best to stick with the official way of loading extention in Keycloak. Just trying to prevent you from shooting yourself in the foot here.

garronej:
Have you check this out: https://docs.keycloakify.dev/testing-your-theme/in-a-keycloak-docker-container  

If it's for developement, using `npx keycloakify start-keycloak` works well and should cover all cases with the options it ussport

PR #693, opened 2024-10-12
Defer script insertion until referenced elements are present in DOM

### LINKED ISSUE #692: WebAuthn button randomly not working
Given the `webauthn-authenticate` page is present, when clicking the “Sign in with Passkey” button, then sometimes nothing happens, and I have to reload the entire page for the button to work.

The console shows

```
TypeError: null is not an object (evaluating 'authButton.addEventListener')
```

This is likely due to the `authenticateWebAuthnButton` element not being present in the DOM when the WebAuthn script is inserted by the `useInsertScriptTags` hook.

marvinruder:
* Fixes #692

### DIFF
--- a/src/login/pages/LoginPasskeysConditionalAuthenticate.tsx
+++ b/src/login/pages/LoginPasskeysConditionalAuthenticate.tsx
@@ -1,4 +1,4 @@
-import { Fragment } from "react";
+import { Fragment, useState } from "react";
 import { clsx } from "keycloakify/tools/clsx";
 import type { PageProps } from "keycloakify/login/pages/PageProps";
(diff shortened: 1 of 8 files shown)


garronej:
Thanks a lot for reporting!  

I though the DOM was guarenteed to be populated when the use effect callback is populated.  
It doesn't seem to be the case.  
I'll fix this in the useScripts hook by observing the DOM so we do not introduce breaking changes for pepole.

garronej:
Thanks you very much for diagnosing this. I've fixed it on every pages that requires scripts.  

@all-contributors please add @marvinruder for bug

PR #701, opened 2024-10-20
Removed Error Message from Terms and Conditions Page

nima70:
In addition to removing the error message on the terms and conditions page, I added a story for long terms and conditions. Since you brought this up, it reminded me that it was also a concern in my own personal project.

image

### DIFF
--- a/stories/login/pages/Terms.stories.tsx
+++ b/stories/login/pages/Terms.stories.tsx
@@ -67,9 +67,6 @@ export const LongMessage: Story = {
     render: () => (
         <KcPageStory
             kcContext={{
-                locale: {
-                    currentLanguageTag: "en"
-                },
                 "x-keycloakify": {
                     messages: {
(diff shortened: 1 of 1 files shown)


stories/login/pages/Terms.stories.tsx
```diff
@@ -45,29 +45,44 @@ export const French: Story = {
...
 };
 
-export const Spanish: Story = {
+export const LongMessage: Story = {
     render: () => (
         <KcPageStory
             kcContext={{
                 locale: {
-                    currentLanguageTag: "es"
+                    currentLanguageTag: "en"
```
garronej, line 71: There's no need to specify the language here since english is the default.  
Let's just specify th properties that diverges from the defaults.

garronej:
Thanks

nima70:
You are welcome...!!!

PR #705, opened 2024-10-24
extended KcContext

jincod:
Hi @garronej

Thank you for this amazing project. 

I'm working on my own keycloak theme. I've made some changes at KcContext. 
- I want to add Social Buttons on Register page
- I want to have a Register link on LoginReset page

### DIFF
--- a/src/login/KcContext/KcContext.ts
+++ b/src/login/KcContext/KcContext.ts
@@ -215,6 +215,19 @@ export declare namespace KcContext {
         recaptchaAction?: string;
         termsAcceptanceRequired?: boolean;
         messageHeader?: string;
+        realm: {
+            password: boolean;
+        };
+        social?: {
+            displayInfo: boolean;
+            providers?: {
+                loginUrl: string;
+                alias: string;
+                providerId: string;
+                displayName: string;
+                iconClasses?: string;
+            }[];
+        };
     };
 
     export type Info = Common & {
@@ -247,6 +260,7 @@ export declare namespace KcContext {
         };
         url: {
             loginResetCredentialsUrl: string;
+            registrationUrl: string;
         };
         auth: {
             attemptedUsername?: string;


garronej:
Hello @jincod,  

In the KcContext type declaration there's only what's used by the default page.  
At runtime there's much more information.  

Keycloakify enables you to extends the type definition of the KcContext.  
See this documentation page: https://docs.keycloakify.dev/faq-and-help/some-values-you-need-are-missing-from-in-kccontext

jincod:
Works perfectly. Thank you!

ddubrava:
> Hello @jincod,
> 
> Thank you for your contribution. In the KcContext type declaration there's only what's used by the default page. At runtime there's much more information.
> 
> Keycloakify enables you to extends the type definition of the KcContext. See this documentation page: https://docs.keycloakify.dev/faq-and-help/some-values-you-need-are-missing-from-in-kccontext

Hey! I'm sorry I didn't want to create a new issue. There are two exported `KcContext` types, one from `kc.gen.tsx` and another one from `KcContext.ts`. What's the purpose of exporting this type from a generated file? From what I understand, I shouldn't use this type in my components

garronej:
Hello @ddubrava,  

It shouldn't be exported from the generated file (`kc.gen.tsx`). Only the KcContext from `src/login/KcContext.ts` should be used.  

`src/login/KcContext.ts` let you declare extra pages that might not be included in the default set of user facing page offered by Keycloak out of the box. Like for example the pages from Phase2 extensions (like [this one](https://github.com/p2-inc/keycloak-magic-link)).  

It also let you declare things on the kcContext type (common to all page or on a page per page basis) that are not used in the default theme but that actually present at runtime.  

Ref:  
https://docs.keycloakify.dev/styling-custom-extension-page  
https://docs.keycloakify.dev/faq-and-help/some-values-you-need-are-missing-from-in-kccontext

ddubrava:
> It shouldn't be exported from the generated file (`kc.gen.tsx`). Only the KcContext from `src/login/KcContext.ts` should be used.

But it's exported - https://github.com/keycloakify/keycloakify-starter/blob/main/src/kc.gen.tsx#L22

We banned the import from this file in ESLint, but I just wanted to clarify.

Thanks

garronej:
> But it's exported  

Yeah well, I tend to export things only not to block usecase that I didn't ancicipated but anyway.  
This type is just an union of the kcContext from the login theme and from the admin theme if you have one (and soon of the admin theme).  

If you have only a login theme importing from `scr/login/KcContext.ts` or `src/kc.gen.tsx` is the same thing. There is just one extra level of indirection if you import it from `src/kc.gen.tsx`.  

I've published an new version where the type is not exported to clear out any confusion.

PR #709, opened 2024-10-29
Draft: Use Story args instead of render()

jirutka:
This is the preferred way in Storybook and it allows us to modify kcContext passed to KcPageStory with Story decorators. This is useful, for example, to add support for globally changing the locale in kcContext via Storybook Toolbar.

https://storybook.js.org/docs/writing-stories/args

I edited the files in bulk using the following two ast-grep rules:

```yaml
id: replace-render-with-args-1
language: TSX
rule:
  all:
    - pattern:
        context: |-
          export const $_VAR = { $STORY }
        selector: object
    - pattern:
        context: "{ render: () => (<KcPageStory kcContext={ $OBJ } />) }"
        selector: object
fix: |-
  {
      args: {
          kcContext: $OBJ
      }
  }
```

```yaml
id: replace-render-with-args-2
language: TSX
rule:
  all:
    - pattern:
        context: |-
          export const $_VAR = { $STORY }
        selector: object
    - pattern:
        context: "{ render: () => <KcPageStory /> }"
        selector: object

fix: "{}"
```

### DIFF
--- a/stories/account/pages/Account.stories.tsx
+++ b/stories/account/pages/Account.stories.tsx
@@ -13,9 +13,7 @@ export default meta;
 
 type Story = StoryObj<typeof meta>;
 
-export const Default: Story = {
-    render: () => <KcPageStory />
-};
+export const Default: Story = {};
 
 /**
  * UsernameNotEditable:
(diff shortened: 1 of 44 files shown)


jirutka:
This cannot be done (yet) because of the outdated storybook version. :(

PR #714, opened 2024-11-06
Fix LoginRecoveryAuthnCodeConfig recovery codes confirm checkbox behavior

thib-rdr:
This pull request fixes the behavior of the recovery codes confirm checkbox in the LoginRecoveryAuthnCodeConfig component. Previously, the checkbox was not functioning correctly, but now it correctly enables and disables the "Save" button based on its checked state.

### DIFF
--- a/src/login/pages/LoginRecoveryAuthnCodeConfig.tsx
+++ b/src/login/pages/LoginRecoveryAuthnCodeConfig.tsx
@@ -1,3 +1,4 @@
+import { useState } from "react";
 import { clsx } from "keycloakify/tools/clsx";
 import { getKcClsx, type KcClsx } from "keycloakify/login/lib/kcClsx";
 import { useScript } from "keycloakify/login/pages/LoginRecoveryAuthnCodeConfig.useScript";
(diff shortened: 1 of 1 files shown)


garronej:
Hello @thib-rdr,  

Thank you very much for diagnosing the problem.  

I can't merge this as is as it would be a breaking change. All the required changes must be done in the useScript hook implementation, since the hook is internalized we can freely modify it.  

I'll do it this week end if I can

garronej:
Hello @thib-rdr,  

After futher review your pr was valid, it could have been merged, it was a stupid error on my part.  

Anyway I opend to implement a more minimal fix to keep the diff minimal: https://github.com/keycloakify/keycloakify/commit/994f1f8d3d00669cb7f564d81e0913f5d32551f1  

Thank you for your contribution!

thib-rdr:
Hi, thanks for your feedback,

If I may, I was wondering if you could consider implementing this change as well : https://github.com/keycloakify/keycloakify/pull/714/files#diff-23fcec1951286d0fe20580506afdd31e66c0782de0923521d70eb6c26b32b4ccR50

As otherwise, the `<p>` tag containing a `<div>` displays an error in the console when the page opens at the moment.

Issue #500, opened 2024-01-31
Customizing emails with React


loschinin:
When customizing emails with React will be available?

garronej:
Hello @loschinin,  
It's not in the imediate priority of the project.

loschinin:
@garronej , Hello and good day. Thank you for your work. Developing email templates in React through keycloakify is like giving developers a superpower. Implementing this would greatly benefit our user engagement and satisfaction. It would be great if you could prioritize this direction.

garronej:
It's definitely a feature that would be nice to have, but there are some points that explain why it's not a priority:

- To support emails, we would need to render React on the server (Keycloak). We can't just ship a JavaScript bundle and render on the client side as we currently do. The server executes FreeMarker templates. While nothing is impossible and there's always a way if there's a will, this definitely presents an engineering challenge. It's interesting, but it means this isn't something I can deliver in a weekend.
- The more HTML and CSS in an email, the more likely it is to end up in the spam folder. Moreover, I believe for emails such as those for resetting passwords or receiving a code, users prefer a simple message like "Here is your code: XXXX" without unnecessary fluff. However, let me know if you think otherwise.
- Emails do not support JavaScript. I'm concerned that users might not understand this and start using things like MUI to build their emails, then blame Keycloakify when it doesn't work as they expected.
- I'm paid to work on [Onyxia](https://onyxia.sh), not Keycloakify. While I'm free to work on any satellite OSS projects that benefit Onyxia, we don't really need email customization, making it hard for me to justify taking a week to work on this. [CloudIAM](https://www.cloud-iam.com/) financially supported the project at one point with around $1500, but they stopped after realizing the links weren't generating any leads. So, until the project gets its own funding, I will prioritize working on features that we need.

Hope this makes sense,

loschinin:
Thank you for your answer.

garronej:
https://youtu.be/Y12sGu8-qFE?si=lpbXqbZrLbQ9NpDf&t=135 might be solving at least some of the issues metioned above

mkreuzmayr:
Also check out https://github.com/shellscape/jsx-email.

garronej:
@mkreuzmayr thanks.  
I think there might be a way to do something here.  
If we prerender the template and search replace for replacing kcContext values with FreeMarker expression.

misterbit-pro:
Hello, is there any ETA on this? Would be great if react email or jsx-email could be used to customize emails

[...]

PR #733, opened 2024-11-22
support themeName prefix in properties for email themes

timofei-iatsenko:
Related https://github.com/keycloakify/keycloakify/issues/500#issuecomment-2490994180

This PR adds a way to override an email subject (and not only) for theme variant. 

Say we have two themes, **vanilla** and **chocolate**. We want to have different email subjects depending on the variant. 

```properties
emailVerificationSubject=Verify email
vanilla.emailVerificationSubject=Verify email for Vanilla
chocolate.emailVerificationSubject=Verify email for Chocolate
```

I decided to do this change because it's small, local, and non-breaking. This will already unblock me, and might be someone else to go with template theming.

Also i was quite surprised that the code mostly doesn't have any tests. I've extracted the emails part and cover it with the tests. This showcasing the approach i use extensively to test CLI in https://github.com/lingui/js-lingui

You can use this approach to test rest of the procedures in the `generateResources`. 

I find testing is very important for maintenance, it's much easier to make changes in the code and just run tests instead of linking it to the starter repo, wait for the full build and so on.

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,5 +121,6 @@
         "webpack-cli": "5.1.4",
         "yauzl": "^2.10.0",
         "zod": "^3.17.10"
-    }
+    },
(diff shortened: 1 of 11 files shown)


timofei-iatsenko:
closing in favour https://github.com/timofei-iatsenko/keycloakify-emails

PR #753, opened 2025-01-06
fix checkbox markup for LoginRecoveryAuthnCodeConfig view

JiiB:
This is a small markup fix to align the checkbox label for the recovery codes. The HTML markup is now the same as the "sign out from other devices" checkbox.

### Screenshot

image

> before (left) after (right)

### DIFF
--- a/src/login/pages/LoginRecoveryAuthnCodeConfig.tsx
+++ b/src/login/pages/LoginRecoveryAuthnCodeConfig.tsx
@@ -65,17 +65,21 @@ export default function LoginRecoveryAuthnCodeConfig(props: PageProps<Extract<Kc
 
             {/* confirmation checkbox */}
             <div className={kcClsx("kcFormOptionsClass")}>
-                <input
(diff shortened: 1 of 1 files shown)


garronej:
Hello @JiiB,  

This is obviously a styling issue but it was directly inherited from Keycloak:  
https://github.com/keycloak/keycloak/blob/8d1956ee5620b49eb2272559ad54f79cce615bd0/themes/src/main/resources/theme/base/login/login-recovery-authn-code-config.ftl#L42-L47  

However in the latest version they seem to have fixed it by hard codding a patternfly class:  

https://github.com/keycloak/keycloak/blob/812d849c783dbee23fcfb7266ce66321665de2bd/themes/src/main/resources/theme/keycloak.v2/login/login-recovery-authn-code-config.ftl#L48  

(which is realy not great because Patternfly are supposed to be strippable but hey... )

Would you try applying the Keycloak patch?

Issue #781, opened 2025-02-04
No translation on title for invitation flow to organisation


acreations:
When you are using Keycloak 26+ and invite a user to an organisation and user gets the invitation email. The link will start Register/Invite already user flow. The title seems to be the key instead of the translated text added under Realm settings.

I think the error is in here

https://github.com/keycloakify/keycloakify/blob/main/src/login/pages/Info.tsx#L23

I tried in my local setup and it seems to be work if I changed this to 

`                        
__html: kcSanitize(messageHeader ? advancedMsgStr(messageHeader) : message.summary)
`

But since Im not the maintainer I not sure what the impact is. Anyway, it is a bug and hopefully it can be fixed (for me Im customising it so I fixed it for our use case)

Thanks for a good extension of Keycloak and made it easier to build an great UI

garronej:
Hello @acreations,  

You're perfectly right, this is an oversight on my end:  

https://github.com/keycloak/keycloak/blob/26.1.1/themes/src/main/resources/theme/base/login/info.ftl

Do you want to submit a PR?

garronej:
(Actually the bug was present in keycloak 24 but they updated it since then https://github.com/keycloak/keycloak/blob/737c21a2559e1c8e2a0e7fdc016c88621f9a00a4/themes/src/main/resources/theme/base/login/info.ftl#L5)

acreations:
Sure I can create a PR for review

acreations:
Added PR https://github.com/keycloakify/keycloakify/pull/782 

I tried follow your guidelines but unfortunately the link ends in a 404

https://docs.keycloakify.dev/faq-and-help/contributing

PR #782, opened 2025-02-05
feat(login): enhance message header translation

acreations:
- Updated Info.tsx to use advancedMsgStr for message header translation.
- This ensures that the message header is properly translated if it exists.

Issue: #781

### DIFF
--- a/src/login/pages/Info.tsx
+++ b/src/login/pages/Info.tsx
@@ -20,7 +20,7 @@ export default function Info(props: PageProps<Extract<KcContext, { pageId: "info
             headerNode={
                 <span
                     dangerouslySetInnerHTML={{
-                        __html: kcSanitize(messageHeader ?? message.summary)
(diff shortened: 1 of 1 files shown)


SpenserJ:
@garronej We ran into this same issue with organization invites (`/realms/realm-name/login-actions/action-token`) and this change fixed the problem. Could you please take a look at it and merge it in for the next release? 🙏

garronej:
Hello, sorry for the late review, this slipped through.  

@coderabbitai what do you think about this fix. Is it consistent with the way we do things?

garronej:
Thank you for the fix @acreations !

garronej:
@allcontributors add @acreations for code

garronej:
@SpenserJ thank you for bringing that to my attention. I've released a new version (patch) that incorporate this code.

PR #793, opened 2025-02-18
feat: allow user press submit button with validation errors from useUserProfileForm hook

EternalSide:
**Enhancement: Flexible Form Validation and Submission Handling in UserProfileFormFields**
**Problem**
Currently, the only way to handle client-side validation and submit the form is through the isFormSubmittable state, which is set in UserProfileFormFields via setIsFormSubmittable. Additionally, the onIsFormSubmittableValueChange prop is mandatory.

The issue with this approach is that it fully disables the submit button when the form is invalid, preventing users from interacting with it. While this behavior may be suitable for some projects, it might not be flexible enough for others.

**Proposed Solution**
To introduce more flexibility, I suggest refactoring the form validation and submission approach by introducing formData.

**Now, instead of relying only on isFormSubmittable, we have a single state object (formData) that allows users to configure form behavior dynamically. The formData object includes:**

mode - active or disabled button
isFormSubmittable – determines whether the submit button should be disabled.
onFormSubmit – a function that handles form submission.

**Enhancements to onFormSubmit**
The onFormSubmit function now accepts:
The form event (event).
An options object as the second parameter, which includes:
shouldSubmit (default: true) – when set to false, it prevents the default form submission, allowing users to handle submission manually after validation.

Depending on the desired behavior, users can adjust formData values to either:
Block the submit button completely (default behavior).
Allow clicking the button, but prevent submission when validation fails, show errors and focus on the first field with error (new behavior).

**Additionally, the UserProfileFormFields component has been updated:**

The onIsFormSubmittableValueChange prop has been removed.
Instead, two new props have been introduced:
mode – determines the form submission behavior.
onFormData– allows managing formState externally.

This update provides greater flexibility while maintaining backward compatibility with the previous validation method.

### DIFF
--- a/src/login/UserProfileFormFields.tsx
+++ b/src/login/UserProfileFormFields.tsx
@@ -15,13 +15,14 @@ import type { KcContext } from "./KcContext";
 import type { I18n } from "./i18n";
 
 export default function UserProfileFormFields(props: UserProfileFormFieldsProps<KcContext, I18n>) {
-    const { kcContext, i18n, kcClsx, onIsFormSubmittableValueChange, doMakeUserConfirmPassword, BeforeField, AfterField } = props;
+    const { kcContext, i18n, kcClsx, doMakeUserConfirmPassword, BeforeField, AfterField, mode, onFormData } = props;
 
     const { advancedMsg } = i18n;
 
     const {
         formState: { formFieldStates, isFormSubmittable },
-        dispatchFormAction
+        dispatchFormAction,
+        onFormSubmit
     } = useUserProfileForm({
         kcContext,
         i18n,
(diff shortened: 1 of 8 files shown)


garronej:
I'm very sorry, but I won't merge this for several reasons:  

- It's a breaking change.  
- It adds 170+ lines of code.  
- It significantly obscures what's actually happening: a POST request is sent to the server on form submission. This might lead people to mistakenly think we're calling an API.  
- It's not abstracted enough. A lot of logic is leaked into the `Register` and `UserProfileFormField` components. These components are userland and must remain as simple as possible. All the logic should be encapsulated within the hooks exposed by the library.  

The only way I'd consider merging this is if the implementation relied on DOM manipulation within `useUserProfileForm` in a framework-agnostic way. However, this would be significantly harder to implement and come with many pitfalls.  

I hope you understand. Thank you for your time and effort—while I won’t merge this upstream, I hope you can still use this code in your theme. Even if it’s not suitable for the library, it remains perfectly valid for application-level code.

PR #841, opened 2025-04-30
use unused variable realm

Jaervinen:
my vite build throws an error for variable realm in line 32 (defined but not used). Therefore, use realm.displayName instead of kcContext.realm.displayName

## Summary by CodeRabbit

- **Refactor**
  - Simplified how the document title is set on the login page for improved code clarity. No changes to user-facing functionality.


garronej:
Thank you :)

Issue #875, opened 2025-06-28
Dutch (nl) translation passed using withCustomTranslations overwrites the English one


BrDewit:
Bug found when using the latest keycloakify version `11.8.49` with keycloak version 26.1.4

In keycloak verson 26.1.4, there seems to not be a translation for the `requiredFields` key as it showed up in the Dutch translation as "Required fields". In the latest keycloak version this translation seems to be added, but as we were still running on that older version, I added the translation to i18n.ts like this:

```
.withCustomTranslations({
    nl: {
        requiredFields: "Vereiste velden"
    }
})
```

This seemed to fix the issue, until I went to look at the English version of the page, and it also said "Vereiste velden", meaning it started showing the Dutch translation.

I fixed this by also adding the English translation to the withCustomTranslations object, but as the requiredFields key exists for the English translation, it of course does not make sense to add it here again, and do the same for potentially every new language added to the keycloak instance.

Download:
Quickly checking your report and I even get a Typescript error when I add the Dutch translation but not the english one:

```console
Property 'requiredFields' is missing in type '{ loginTitleHtml: string; }' but required in type 'Record<"loginTitleHtml" | "requiredFields", string | Record<"keycloakify-starter", string>>'.ts(2741)
```

garronej:
Yes sorry guys, thank you for reporting.  
There is a few of thoses little issues arising. I think the priority for me should be to work on login v2 rather than trying to fix all of them.

BrDewit:
@garronej I completely understand. I've just looked at the code myself and found that the issue occurs due to the language fallback mechanisms on line 222 - 231 of getI18n.tsx

I'll create a pull request with the necessary changes.

PR #876, opened 2025-07-03
Fix for issue #875: Custom translations for one language showing for other languages

BrDewit:
In issue #875 I outlined an issue where creating a custom translation for one language, caused all other languages to display that custom translation, instead of only the specified language displaying the custom translation.

The cause of the issue was due to the fact that the code fell back on the custom translations of other languages when creating the translation function of a specific language. If there was no custom translation for the language at all, for any key, the custom translation of either the default language, or if that didn't exist, the first specified custom translation, was used.
If any custom translation existed for a certain language, the issue would not occur as the `messagesByLanguageTag_themeDefined[currentLanguage.languageTag]` call on line 223 of the getI18n.tsx file does not return undefined in this case, and it would not fall back on another language.

I've removed the fallback to other languages in this part of the code, as the custom translation is supposed to only be applied to the language it's specified for. The resolveMsg and resolveMsgAdvanced functions handle falling back to other languages if a certain key does not exist.

I've tested the effects of the changes manually by testing multiple configurations of different languages and translations, but I've not implemented an automated test.

The change is only breaking if the user of the library used a custom translation of an existing key in 1 language to translate that same key across multiple/all other languages.

## Summary by CodeRabbit

* **New Features**
  * Improved translation fallback on the login screens: theme-specific text now falls back to a secondary language when missing, reducing missing or incorrect text for users.

* **Refactor**
  * Streamlined message resolution to more reliably select theme and fallback translations.

* **Bug Fixes**
  * Fewer missing-translation issues and more consistent localized UI.


garronej:
Hello @BrDewit,  

Thank you for the well detailed PR.  

That being said I think this changes, while fixing the issue of overwriting the curstom translation of **an existing key** in only a specific language this change also cause important regression when defining custom message key.  

```ts
.withCustomTranslations({
    en: {
        myCustomMessage: "Hello world"
    }
})
```

With this PR applied, if the current language is `en`, `msgStr("myCustomMessage")` gives `"Hello World"` but if the current language is, let's say, `fr` then it gives "myCustomMessage".  

Defining custom, non standard message keys is the primary usecase of `withCustomTranslations`.  

What would be a reasonable approach is, if the key is a known one then do what you suggest otherwise keep the current behavior....  

But even then, you usually never need to overwrite one of the default message keys, in your case you do just because the translation file are outdated and I have to update them.  
But in the general case, if you have overwriten a default message it's probably because you've altered the sementic and want this new semantic to be always visible, even in the wrong language.  

This is debatable though but as it stands the rule of `withCustomTranslations` is. If you want to use it, you must provide a translation for every language you wish to support or we will fall back to english or the first one.

BrDewit:
Hi @garronej

That's a very reasonable remark, if they did not specify the custom message, you would indeed want to fall back to an existing translation as that's better than just having the key inserted in that position. Especially as that's probably a more common case than the one where a custom translation is given for an existing key. When I have time, I'll revise the changes to fix this regression.

BrDewit:
@garronej excuse me for the delay, but I've finally made the change.

The translation logic will now first check in the custom translation of the provided language, then it will check the default set of the provided language, then it will check the custom translation of the fallback language, after which it will check the default set of the fallback language.

This way, my issue, where a custom translation in 1 language would override the default translation of that key in every language, in is fixed.
While the regression is also fixed, which would cause the key itself to be shown instead of a translation, when no custom translation of a non-default key was provided in the language, but it was in another.

The only breaking change is of course if someone used the behaviour I changed, as a feature, where they use one custom translation to override default keys for every language, but that behaviour is unintuitive. It seems reasonable that with `withCustomTranslations` you must provide a translation for the key in every language you wish to override.

BrDewit:
Hi @garronej I recently remembered that my pull request was still open. I've updated my fork with the latest changes, and the changes I made still seem relevant. If you agree with the changes, the pull request can be merged. If not, let me know what needs to be changed!

Thanks in advance and happy holidays :)

Issue #877, opened 2025-07-03
Buttons on update email form


DerIch27:
The update email form has two buttons if isAppInitiatedAction is true and one otherwise. I assume that the first button should have the class kcButtonBlockClass if it is the only button and don't have the class otherwise (like keycloak does is). For some reason it seams to be implemented the other way around by keycloakify.

The keycloak default theme presents the following form
![Image](https://github.com/user-attachments/assets/9053d676-4dad-4305-9d09-0b5158cd9558)
while keycloakify's default theme presents the following form
![Image](https://github.com/user-attachments/assets/7480a1fd-3f58-4ddc-bf15-349b70f2f654)

Thanks for your work!

garronej:
Hello @DerIch27, thank you for reporting. There is probably a little implementation error here. PR wellcome!

PR #878, opened 2025-07-04
Fix button styles on update email and login expired forms

DerIch27:
solves issue #877 

The submit button on the update mail form should have the class block using the whole width if it is the only button in the form and be small if the form has an additional cancel button next to ist.

I added a space at the login expired page before the link because it separates two words and gives the same style like in the line below.

## Summary by CodeRabbit

* **Style**
  * Improved spacing consistency in the login page by adding a trailing space after a link.
  * Updated button styling on the email update page for better visual consistency.


garronej:
Good catch: https://github.com/keycloak/keycloak/blob/8ce8a4ba089eef25a0e01f58e09890399477b9ef/themes/src/main/resources/theme/base/login/update-email.ftl#L24  

Thank you for the PR!

garronej:
@allcontributors add @DerIch27 for code

PR #904, opened 2025-08-09
Refactor package manager lock file handling to support multiple lock …

0-don:
…file names

### DIFF
--- a/src/bin/tools/npmInstall.ts
+++ b/src/bin/tools/npmInstall.ts
@@ -18,34 +18,34 @@ export async function npmInstall(params: { packageJsonDirPath: string }) {
         const packageMangers = [
             {
                 binName: "yarn",
-                lockFileBasename: "yarn.lock"
+                lockFileBasenames: ["yarn.lock"]
             },
             {
                 binName: "npm",
-                lockFileBasename: "package-lock.json"
+                lockFileBasenames: ["package-lock.json"]
             },
             {
                 binName: "pnpm",
-                lockFileBasename: "pnpm-lock.yaml"
+                lockFileBasenames: ["pnpm-lock.yaml"]
             },
             {
                 binName: "bun",
-                lockFileBasename: "bun.lockdb"
+                lockFileBasenames: ["bun.lock", "bun.lockb", "bun.lockdb"]
             },
             {
                 binName: "deno",
-                lockFileBasename: "deno.lock"
+                lockFileBasenames: ["deno.lock"]
             }
         ] as const;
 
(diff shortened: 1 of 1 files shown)


garronej:
Hey @0-don,  

Thank you for the PR.  
I few weeks ago that would have been a hard no but now I'm seriously considering it since the NPM CLI is now trhowing a ton of irrellevent warnings. [Ref](https://github.com/npm/cli/issues/8353).  

It looks good on paper, I'm just very afraid it will break things for pepole and I'm not looking forward to runtime specific issues.  

I'll review it in the next few days.

garronej:
Hey @0-don,  

Sorry if my previous reply came across as a bit cold, your contribution is genuinely valuable.  

A few points to consider: `npx xx` is essentially just shorthand for “look in `node_modules/.bin/xx` and run it with Node.” It’s not doing any real work; it’s just a small convenience wrapper. That’s why it’s frustrating that NPM now throws warnings — we shouldn’t have to figure out which package manager a user is using just to run an installed JavaScript script. But because of those warnings, we’re forced to either swallow them or do what you suggest.  

That said, your approach goes further — you’re proposing switching the **JavaScript runtime** itself. This makes sense if you’re using `bun` or `deno` instead of `node` as the runtime, because when you run a Keycloakify command you’d expect the runtime to handle everything. Currently, though, Node and NPM still need to be installed, and Node is actually spawned for part of the work.  

So I just wanted to explain why I need to proceed carefully here and run some tests before merging your changes.

0-don:
yeah, no problem take your time

0-don:
https://github.com/antfu-collective/ni here is a good example, maybe this helps

garronej:
Thank you @0-don!  

This is **very** interesting indeed.  
My first reaction was: *“Great, we can vendor it as-is and finally put this issue to rest.”*  

But when I ran a quick check on one of the most common errors in this area, it failed right away:  

image

That’s a problem. `npx vitest` and `pnpm dlx vitest` behave **completely differently**:  
- `pnpm dlx vitest` always installs to an ephemeral sandbox (separate from your project) and runs from there.  
- `npx vitest` will use the local binary if Vitest is installed in your project, otherwise it fetches a temporary version from npm.  

This discrepancy is something I run into often—badly enough that I had to hard-code a guard for it:  

image

So while the implementation looks very promising, I need to be extremely cautious before adopting it.  
If they got something this fundamental wrong, there are probably other pitfalls waiting.  

Still, I really appreciate you sharing it—thank you again!

PR #919, opened 2025-09-04
feat: ability to skip jar output

zapkub:
Introduce a new flag for build command `--skip-jar` to skip output as jar file and leave the pre-bundle folder.

I would like to submit this change because We use a different tools to bundle JAR output in our build system.

This will allow little flexibility to bundle the final output.

## Summary by CodeRabbit

- New Features
  - Added a --skip-jar option to the build command to skip generating the output JAR.
  - When used, Maven presence checks are bypassed, enabling builds in environments without Maven.
  - Skipping the JAR can speed up development builds; default behavior remains unchanged (JAR is built unless --skip-jar is specified).
  - Clear warning is logged when JAR generation is skipped.


zapkub:
Close and replace by https://github.com/keycloakify/keycloakify/pull/921

PR #921, opened 2025-09-08
feat: support predefined context and skip postbuild / skip jars

zapkub:
Support build in non `npm, npx` build system by introduce several options to run build command to avoid `child_process` execution.

Mostly the subprocesses are used inside build context factory function. This PR introduce a predefined result of the build context factory function.

- Support predefined build context via env var `KEYCLOAKIFY_PREDEFINED_BUILD_CONTEXT`
- Add `--skip-post-build` option to skip post build sub process
- Add `--skip-jar` option to skip mvn packing

## Test

I have tested with the script below to make sure external sub process won't be
invoked during the build

```
#!/bin/bash

set -e

export KEYCLOAKIFY_PREDEFINED_BUILD_CONTEXT=$(./node_modules/.bin/keycloakify print-build-context)

npx() {
    echo "npx is forbidden"
    exit 1
}
export -f npx

npm() {
    echo "npm is forbidden"
    exit 1
}
export -f npm

mvn() {
    echo "mvn is forbidden"
    exit 1
}
export -f mvn

./node_modules/.bin/vite build
./node_modules/.bin/keycloakify build --skip-jar --skip-post-build
```

## Summary by CodeRabbit

- New Features
  - Added CLI flags: --skip-jar to skip JAR generation and related checks, and --skip-post-build to skip post-build steps.
  - Introduced print-build-context command to output the project's build context as JSON for external build runners.
  - Added support for KEYCLOAKIFY_PREDEFINED_BUILD_CONTEXT env var to preload a build context without computing it.
  - Improved user-facing logs, clearly indicating when JAR building or post-build steps are skipped.

### DIFF
--- a/src/bin/keycloakify/keycloakify.ts
+++ b/src/bin/keycloakify/keycloakify.ts
@@ -10,10 +10,18 @@ import { readThisNpmPackageVersion } from "../tools/readThisNpmPackageVersion";
 import * as os from "os";
 import { rmSync } from "../tools/fs.rmSync";
 
-export async function command(params: { buildContext: BuildContext }) {
+export async function command(params: {
+    buildContext: BuildContext;
+    skipJar?: boolean;
+    skipPostBuild: boolean;
+}) {
     const { buildContext } = params;
 
     exit_if_maven_not_installed: {
+        if (params.skipJar) {
+            break exit_if_maven_not_installed;
+        }
+
         let commandOutput: Buffer | undefined = undefined;
 
         try {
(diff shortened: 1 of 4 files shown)


garronej:
Hello @zapkub,  

First, thank you for taking the time to put together this PR, I really appreciate the effort and the thought behind it. 🙏  

That said, I don’t think we should move forward with merging this change.  

The problem this PR addresses (preventing subprocess execution in CI environments) comes from very specific build policies. I completely understand where that comes from, many organizations have strict compliance or reproducibility requirements.  
However, from Keycloakify’s perspective, introducing additional flags like `--skip-jar`, `--skip-post-build`, or environment-driven build contexts adds a lot of complexity to the build pipeline for a very niche use case.  

The packaging process is already one of the most intricate parts of the project, and every new code path or flag increases maintenance costs and potential edge cases. Given that, I’d prefer to keep the CLI focused and stable, rather than expanding it to accommodate custom CI rules.  

If you need to adapt Keycloakify to your environment, I’d recommend using [`patch-package`](https://www.npmjs.com/package/patch-package).  
It’s a tool I personally use when I need to slightly adjust a dependency to fit a very specific constraint. It lets you maintain a small local diff that’s automatically applied after each install, no forking or waiting for upstream changes, and you stay in full control of your modification.  

So again, thank you for the contribution and for taking the initiative to make Keycloakify work better in your setup. I really appreciate the effort, I just think this particular customization is better handled downstream. 🙏

zapkub:
@garronej Thank you for your response. We will handle the change in our side and looking forward to contribute the build feature in the future.

PR #922, opened 2025-09-12
Fix warning related to hostname v1 feature

rgrosjean:
Fix a warning on startup after deprecation of hostname_v1:

Keycloak 25
        - hostname-strict-https: Available only when hostname:v1 feature is enabled.

Keycloak 26
WARNING: Hostname v1 options [hostname-strict-https] are still in use, please review your configuration

## Summary by CodeRabbit

* **Bug Fixes**
  * Keycloak startup now conditionally includes the HTTPS hostname strictness setting based on the detected major version: it is retained for versions below 25 and omitted for 25 and above. This preserves prior behavior for older releases while improving compatibility and reducing configuration friction with newer Keycloak versions, resulting in smoother, more reliable startups.


garronej:
Thank you!

PR #929, opened 2025-10-02
Allow target="_blank" on <a> elements in KcSanitizer

### LINKED ISSUE #763: Allow "target" property in anchors on KcSanitizer
I'd like to open the subscription terms and the privacy policy in a new tab instead of opening it in the same tab.

I added the following to the i18n.ts file:
```
           termsText:
                'Please accept the <a href="https://link.to/subscription-terms" target="_blank">subscription terms</a>.',
```

and call it like this:
```
          <label htmlFor="termsAccepted" className={kcClsx("kcLabelClass")}>
             {msg("acceptTerms")}
          </label>
```

I found that the i18n string is sanitized using the `kcSanitize` function. There, it actually removes the `target="_blank"` part of the link. Could this be added or is there another way to achieve this desired functionality?

kodebach:
Fixes #763 by adapting the upstream PR https://github.com/keycloak/keycloak/pull/42700

## Summary by CodeRabbit

- New Features
  - Links in sanitized content can now open in a new tab via target="_blank".
  - When target="_blank" is used, rel="nofollow" is automatically applied for safer linking.
  - Other target values remain unchanged and are not modified with rel attributes.

- Tests
  - Added coverage confirming target="_blank" is preserved and paired with rel="nofollow".
  - Verified links without target or with other target values do not receive rel attributes.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


Patrizio:
@garronej We would be very happy to see this fix or suggestion in the next version of Keycloakify. How do you see the chances there?

garronej:
Thank you @kodebach!

garronej:
@Patrizio I've just released a new version that incorporate this fix.

Patrizio:
@garronej Thank you 💯

Issue #560, opened 2024-05-29
Error: Command failed: mvn clean install


Aldiuser:
Hello everyone, 

**I hope someone knows a solution. If you need more information, I can provide it!**

I am trying to build a Keycloak theme with my pipeline in openshift (tekton Pipelines) and get the following error message. I only get this error message in an environment with a proxy. If no proxy is activated, access works properly and the Keycloak theme is created. I have set up my proxy in NPM and discovered in your code that you read this information from NPM, so there is only this one place where I set the proxy in the script:

```
npm config set proxy ${http_proxy}
npm config set https-proxy ${https_proxy}
npm config set noproxy ${no_proxy}
```

**Environment Informations:**

```
keycloakify: 9.7.0

I use the image node:18 

Apache Maven 3.8.8 (4c87b05d9aedce574290d1acc98575ed5eb6cd39)
Maven home: /workspace/working-directory/themes/pos-theme/apache-maven-3.8.8
Java version: 17.0.11, vendor: Debian, runtime: /usr/lib/jvm/java-17-openjdk-amd64
Default locale: en_US, platform encoding: ANSI_X3.4-1968
OS name: "linux", version: "5.14.0-284.52.1.el9_2.x86_64", arch: "amd64", family: "unix"

Node.js v18.20.3
```

**Error Message:**

```
> keycloakify-starter@6.1.9 build
> npm run checkVersion && npm run build-vite && keycloakify


> keycloakify-starter@6.1.9 checkVersion
> pwd

/workspace/working-directory/themes/pos-theme

> keycloakify-starter@6.1.9 build-vite
> tsc && vite build -d

2024-05-29T08:23:42.805Z vite:config bundled config file loaded in 299.40ms
2024-05-29T08:23:47.549Z vite:config using resolved config: {
  plugins: [
    'vite:build-metadata',
    'vite:watch-package-data',
    'vite:pre-alias',
    'alias',
    'vite:react-babel',
    'vite:react-refresh',
    'vite:modulepreload-polyfill',
    'vite:resolve',
    'vite:html-inline-proxy',
    'vite:css',
    'vite:esbuild',
    'vite:json',
    'vite:wasm-helper',
    'vite:worker',
    'vite:asset',
    'vite-plugin-commonjs',
    'keycloakify',
    'vite:wasm-fallback',
    'vite:define',
    'vite:css-post',
    'vite:build-html',
    'vite:worker-import-meta-url',
    'vite:asset-import-meta-url',
[...]

PR #934, opened 2025-10-10
feat: add option to use default local maven repo

emouty:
when building on our CI we are required to fetch our dependencies from our Google cloud artifactregistry. 

To log onto it so we are using `com.google.cloud.artifactregistry:artifactregistry-maven-wagon`. since the current maven command use option `maven.repo.local` it can't access the dependency jar that is present in the default

```
 [ERROR] Error executing Maven.
 [ERROR] Extension com.google.cloud.artifactregistry:artifactregistry-maven-wagon:2.2.5 or one of its dependencies could not be resolved: Plugin com.google.cloud.artifactregistry:artifactregistry-maven-wagon:2.2.5 or one of its dependencies could not be resolved:
 	Failed to read artifact descriptor for com.google.cloud.artifactregistry:artifactregistry-maven-wagon:jar:2.2.5
 
 [ERROR] Caused by: Plugin com.google.cloud.artifactregistry:artifactregistry-maven-wagon:2.2.5 or one of its dependencies could not be resolved:
 	Failed to read artifact descriptor for com.google.cloud.artifactregistry:artifactregistry-maven-wagon:jar:2.2.5
```

I tried to copy the artifact to the local repo but since there is a clean-up of the repo ran within during the build command execution. I can't find a way to provide it to keycloakify. 

In order to allow user to log onto registry using mvn extensions I have added an options to use the default maven repo instead of a custom one 

## Summary by CodeRabbit

* **New Features**
  * Build process can optionally use the default Maven repository or a local cache, allowing more flexible dependency resolution via environment configuration.

* **Refactor**
  * Internal build context and configuration handling refined for more robust and maintainable build orchestration.

* **Chores**
  * Minor code organization and import ordering improvements.

### DIFF
--- a/src/bin/keycloakify/buildJars/buildJar.ts
+++ b/src/bin/keycloakify/buildJars/buildJar.ts
@@ -3,13 +3,13 @@ import type {
     KeycloakAccountV1Version,
     KeycloakThemeAdditionalInfoExtensionVersion
 } from "./extensionVersions";
-import { join as pathJoin, dirname as pathDirname } from "path";
+import { dirname as pathDirname, join as pathJoin } from "path";
 import { transformCodebase } from "../../tools/transformCodebase";
 import type { BuildContext } from "../../shared/buildContext";
 import * as fs from "fs/promises";
 import {
-    generatePom,
-    BuildContextLike as BuildContextLike_generatePom
+    BuildContextLike as BuildContextLike_generatePom,
+    generatePom
 } from "./generatePom";
 import { readFileSync } from "fs";
 import { isInside } from "../../tools/isInside";
@@ -222,7 +222,14 @@ export async function buildJar(params: {
     }
 
     {
-        const mvnBuildCmd = `mvn -B -ntp clean install -Dmaven.repo.local="${pathJoin(keycloakifyBuildCacheDirPath, ".m2")}"`;
+        const useDefaultMavenRepo =
+            process.env.KEYCLOAKIFY_USE_DEFAULT_MAVEN_REPO === "true";
+
+        const mavenLocalRepoArg = useDefaultMavenRepo
+            ? ""
+            : `-Dmaven.repo.local="${pathJoin(buildContext.cacheDirPath, ".m2")}"`;
+
+        const mvnBuildCmd = `mvn -B -ntp clean install ${mavenLocalRepoArg}`.trim();
 
         await new Promise<void>((resolve, reject) =>
             child_process.exec(


emouty:
Hello @garronej , any chances for you to get a look at this pr  ?

garronej:
@coderabbitai I'm not understanding what's going on precisely.  
What I do is instead of just running `mvn clean install` I will install with a -Dmaven.repo.local= argument pointing to a dir within the keycloakify cache reserved dir. 
For context, in theory we shouldn' have to download any JAR extention but we do to poliffil some features present in older keycloak versions.  
But it's a pain since it's creating issue for pepole behind corporate proxies.  

I use -Dmaven.repo.local for two reasons:
1) I don't want `npx keycloakify build` to generate any files in random directory. I want the process to be safe and reproducible.  
2) I want user to avoid re-downloading the dependency every time the theme is build, with the .m2 dir being in the cache, the dependencies are downloaded only once.  
3) It enables users that would have defined [XDG_CACHE_HOME](https://docs.keycloakify.dev/features/compiler-options/xdg_cache_home) to pre build their theme on unrestricted network then to reaply this dir in the CI so they can build offline.  

Now what do you think about this PR? As far as I know other users have been able to customize proxy and other stuff by adding some other M2 to the path. See: https://github.com/keycloakify/keycloakify/issues/560#issuecomment-2834853618  
Can you clarify for me?

garronej:
@coderabbitai  

> 	•	It's opt-in (defaults to false, preserving current behavior)  

Yeah but it's also a knob and quite an obscure one. I try to avoid letting user add their little options because it can quickly spiral out of control making, increacing the API surface when the problem could have been handled internally.  

What I don't understand is why the user need extentions in the first place, if you look at the code you can see that what we're doing is basically use maven to create a zip of a ton of static assets plus in some case, add a couple of dependency for jar targetting older keycloak version (or for the account v1 admin console that has been removed from keycloak).  
As far as I remember this is not customizable. 
So why do the user need extention?

garronej:
@coderabbitai ok that makes sense.  

The suggestions of alternative implementations are also good but it's too much of an oportunity cost.  

I want to solve the problem for this user though. So I'm considering two options:  

1) Merge this PR as is. Adding some noise to the API and introducting an option that silently break build reproductibility.  
2) Make the user update their PR to use an ENV variable instead of a CLI option for implementing heir workaround.  

The reasoning is that workaround that are not fully satisfactory should not be documented, and remain hidden. 
They are something that I can share with pepole having the same kind of issue. 
If it's ask often enough it can be upgraded into a documented knob or the problem can be tackled for real, but not before.  

So what do you think?

garronej:
@coderabbitai I think it's best not to make it leak to the buildContext and just use the env in the buildJar.ts

@emouty Thank you for the PR! I'll happily merge a rework of this PR with an undocumented special Env variable, see last message.

emouty:
I have edited the PR to introduce the hidden env var `KEYCLOAKIFY_USE_DEFAULT_MAVEN_REPO` instead of using a cli option :)

garronej:
@allcontributors add @emouty for code

Issue #881, opened 2025-07-07
Sanitizer not allowing `<br />` tag


AlexLeoTW:
## What Happened?

I configure my realm with following "HTML Display name".

```html
my super duper looooooong <br></br> realm title
```

(Ps. I also tried `<br>`, `<br />`)

And this rendered in the browser.

```html
<div id="kc-header-wrapper" class="kcHeaderWrapperClass">
  <span data-kc-msg="loginTitleHtml">
    my super duper looooooong  realm title
  </span>
</div>
```

Printing the values in the console, and I got the following

```js
console.log(realm.displayNameHtml) // my super duper looooooong <br> realm title
console.log(kcSanitize(realm.displayNameHtml)) // my super duper looooooong  realm title

console.log(JSON.stringify(msg("loginTitleHtml", realm.displayNameHtml))); 
// {"type":"span","key":null,"ref":null,"props":{"data-kc-msg":"loginTitleHtml","dangerouslySetInnerHTML":{"__html":"my super duper looooooong  realm title"}},"_owner":null}
```

## How to Reproduce ?

Run a Keycloak container with them loaded, then configure "HTML Display name" with something containing `<br />` tag

<img width="2133" height="1315" alt="Image" src="https://github.com/user-attachments/assets/4a2ad4dc-593b-4cda-b3ec-8a8f85b40055" />

PR #940, opened 2025-10-21
#881: allow <br> tags

rgrosjean:
## Summary by CodeRabbit

* **Bug Fixes**
  * HTML sanitization now treats tag names case-insensitively and restricts automatic removal to a defined set of simple formatting tags, preventing unintended stripping and better preserving surrounding content.

* **Tests**
  * Added tests to verify preservation of break tags, horizontal rules, and adjacent text, and to ensure non-whitelisted tags without attributes are removed while keeping their inner text.

### DIFF
--- a/src/lib/kcSanitize/HtmlPolicyBuilder.ts
+++ b/src/lib/kcSanitize/HtmlPolicyBuilder.ts
@@ -34,7 +34,7 @@ export class HtmlPolicyBuilder {
     }
 
     allowWithoutAttributes(tag: string): this {
-        this.tagsAllowedWithNoAttribute.add(tag.toLocaleLowerCase());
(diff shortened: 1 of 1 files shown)


rgrosjean:
hi @garronej I based this change on the OWASP Java HTML Sanitizer documentation:
https://github.com/OWASP/java-html-sanitizer?tab=readme-ov-file#custom-policies

> Please note that the elements "a", "font", "img", "input" and "span" need to be explicitly whitelisted using the allowWithoutAttributes() method if you want them to be allowed through the filter when these elements do not include any attributes.

garronej:
@coderabbitai This seems like a faily straight forward and wellcome addition. 
Make the argument for and against merging it. Conclude with your recomendation.

garronej:
Than you @rgrosjean!  

@allcontributors add @rgrosjean for code and test

Issue #944, opened 2025-10-29
The "Back to Application" button’s href is not set when no Home URL is provided.


fileng87:
Im current working on version 26.4.2

Keycloak setting like this

<img width="1294" height="831" alt="Image" src="https://github.com/user-attachments/assets/4b3915e0-c68c-4337-8a0e-90c75cb312f1" />

The "Back to Application" button’s href is not set, maybe it should be hidden

<img width="667" height="553" alt="Image" src="https://github.com/user-attachments/assets/f3b222bb-62e7-4774-9352-b0d3d4a6608c" />

garronej:
Hello @fileng87,  

It doesn't work like this. Think about it, the account console is shared for all the apps in the realms.  
If you define on a specific client a home url, how would keycloak know form the admin console to wich app you want to go back to? It can't.  
This is why, from your application, when you redirect to the account page, you have to provide the `&referrer=<client-id>` parameter to the url and also the refferer_uri.  

https://github.com/user-attachments/assets/a1766eab-f1d4-4fcb-9d59-8f9c1e180d1e   

If you are not using Next.js you can checkout https://www.oidc-spa.dev/ if you want a clean integration of Keycloak in your application.

fileng87:
Oh, I see now. I thought it would fall back to the root URL, but if I use the default theme and don’t provide a Home URL, the “Back to Application” button won’t be displayed.

<img width="1879" height="934" alt="Image" src="https://github.com/user-attachments/assets/4b6cbc9f-a5a1-42f2-971c-4b341c2ebc2b" />

Sorry, my description was poor — this happens on the login error screen.

fileng87:
I found the issue, in the [Error.tsx](https://github.com/keycloakify/keycloakify/blob/main/src/login/pages/Error.tsx) line 24. 

The client.baseUrl can be an empty string, which causes the "Back to Application" button to display on the login error page with an empty href when home url doesn't provide.

I fixed this bug by changing:
```ts
!skipLink && client !== undefined && client.baseUrl !== undefined
``` 
to 
```ts
!skipLink && client !== undefined && client.baseUrl
```

garronej:
@fileng87 Good, thank you! Would you want to submit a PR for that?

PR #948, opened 2025-10-31
Fix “Back to Application” button showing on login error page when Home URL is not set

fileng87:
The client.baseUrl can be an empty string, which causes the "Back to Application" button to display on the login error page with an empty href when home url doesn't provide.

fix by changing:
```ts
!skipLink && client !== undefined && client.baseUrl !== undefined
``` 
to 
```ts
!skipLink && client !== undefined && client.baseUrl
```

issue #944 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the logic for displaying the back-to-application link on error pages, ensuring it only appears when a valid application URL is configured.


fileng87:
Looks like this was already fixed in #1009, so I’m closing this PR for now.

PR #1025, opened 2026-06-03
Fix DEP0190 spawn deprecation warnings introduced in node 24.15

githrdw:
Node introduced these deprecation warnings when passing arguments seperately into a spawn combined with `{ shell: true }`:
https://nodejs.org/api/deprecations.html#dep0190-passing-args-to-nodechild_process-execfilespawn-with-shell-option

This change merges the command and arguments into a single param to omit the deprecation warning.

## Summary by CodeRabbit

* **Chores**
  * Refactored internal build and development server initialization to use simplified process spawning approach across multiple development scripts.

**Note:** These are infrastructure-level changes with no visible impact to end-user functionality.

### DIFF
--- a/scripts/shared/startRebuildOnSrcChange.ts
+++ b/scripts/shared/startRebuildOnSrcChange.ts
@@ -13,7 +13,7 @@ export function startRebuildOnSrcChange() {
 
         const dCompleted = new Deferred<void>();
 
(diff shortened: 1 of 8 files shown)


garronej:
@allcontributors please add @githrdw for code.

garronej:
Thank you for your contribution @githrdw!

PR #1033, opened 2026-06-10
[Security] Bump DomPurify package

### LINKED ISSUE #1031: Keycloakify ships a vulnerable version of DOMPurify
The latest Keycloakify ships with a version of `DOMPurify` that has known high severity vulnerabilities as shown [here](https://nvd.nist.gov/vuln/detail/CVE-2025-15599). The current version of DOMPurify that is included in the built assets is 3.1.6 (as installed [here](https://github.com/keycloakify/keycloakify/blob/main/package.json#L96)).

Since this dependency is inlined in the vendor folders of this repo, we should ideally bump it to 3.4.8 which is the latest version and free of any security considerations.

I'm happy to own the work and put up a PR if this is something that the maintainers would be ok to approve/merge.

Thanks!

3nvi:
## Description

The latest Keycloakify ships with a version of DOMPurify that has known high severity vulnerabilities as shown [here](https://nvd.nist.gov/vuln/detail/CVE-2025-15599). The existing version of DOMPurify that is included in the built assets is 3.1.6 (as installed [here](https://github.com/keycloakify/keycloakify/blob/main/package.json#L96)).

In this PR we are bumping it to the latest  `v3.4.9` to fix all the related security vulnerabilities. While there, we also bump the isomorphic version of this to fully evict the vulnerable version from the lockfile.

This PR closes https://github.com/keycloakify/keycloakify/issues/1031


3nvi:
@garronej given the sec vuln & the fact that this is an package related to authentication, the sooner we can merge it the better 😉 

P.S. I've pnpm patched it in our own app and it worked nicely

garronej:
Thank you!

garronej:
@3nvi Just a bit of context to explain why I wasn't more proactive about this matter.

There is no actual security risk here.

DOMPurify is effectively useless in Keycloakify. We use it to sanitize HTML strings that have already been sanitized upstream or that are not user-generated in the first place.

So why did I use it?

Because I eventually gave in. People would see <div dangerouslySetInnerHTML={template} /> and immediately say, "This is unsafe; the template should be sanitized."

It's difficult to explain how I can guarantee that this is safe in a way that convinces people who aren't familiar with Keycloakify's internal architecture. Rather than having that discussion over and over again, I accepted introducing DOMPurify.

PR #1034, opened 2026-06-18
fix: add missing totp username property

whawker:
A quick fix to add the missing `username` property to the TypeScript definitions of the `login-config-totp` page

This is provided by Keycloak [here](https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/forms/login/freemarker/model/TotpBean.java#L121-L123), I suspect it was missed because the current FTL template doesn't make use of it.

I would like this property present so I can build a `otpauth://` in addition to the QR code without having to use ts-ignore. 

## Summary by CodeRabbit

* **New Features**
  * Enhanced TOTP authentication flows to include username support across account and login contexts, improving credential handling during two-factor authentication setup and configuration.

### DIFF
--- a/src/account/KcContext/kcContextMocks.ts
+++ b/src/account/KcContext/kcContextMocks.ts
@@ -173,7 +173,8 @@ export const kcContextMocks: KcContext[] = [
                 type: "totp",
                 period: 30,
                 getAlgorithmKey: () => "SHA1"
-            }
+            },
+            username: "alice@example.com"
         },
         mode: "qr",
         isAppInitiatedAction: false,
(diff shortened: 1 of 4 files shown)


src/account/KcContext/kcContextMocks.ts
```diff
@@ -173,7 +173,8 @@ export const kcContextMocks: KcContext[] = [
                 type: "totp",
                 period: 30,
                 getAlgorithmKey: () => "SHA1"
-            }
+            },
+            username: "alice@example.com"
```
whawker, line 177: This is the same email address as used in the Storybook QR code

garronej:
Thank you @whawker!

garronej:
Would you open a PR for here as well https://github.com/keycloakify/account-multi-page-ui ?

whawker:
Sure! Opened https://github.com/keycloakify/account-multi-page-ui/pull/2

garronej:
@allcontributors add @whawker for code

Issue #1042, opened 2026-07-22
[question] Expose generate resources api


nrakochy:
Hi there, thanks for the great library. I am hoping to use keycloakify for our system and hitting a bit of snafu when adding it into our CI pipeline. Specifically, the `keycloakify` executable has a few different build assumptions that make it tricky to integrate in our specific (JS) env. 

Would you be open to a PR that exposes the `generateResources` function? This is very similar to #647, with the slight caveat that it would be great to be able to invoke the function from the lib in node rather than as a cli cmd, but I believe either could work. Happy to contribute 🙏

garronej:
See [this answer](https://github.com/keycloakify/keycloakify/pull/1048#issuecomment-5152338262)

PR #1048, opened 2026-07-29
feat: Expose generateResources as a public Node API

nrakochy:
Implementation of #1042. Happy to change as needed, please let me know if there's a shape of this that would work.

I added a `keycloakify/api` entry point exposing `generateResources()`, the resource generation half of `keycloakify build`, without the Maven jar packaging step. This makes the theme resources buildable from a JS build pipeline, or from a build system other than Maven.

## Usage

```ts
import { generateResources } from "keycloakify/api";

// The bundler build must run first — this reads its output, it doesn't produce it.
// $ npx vite build

const { resourcesDirPath } = await generateResources({
  resourcesDirPath: "build_keycloak/resources", // required; deleted + recreated
  projectDirPath: process.cwd(), // optional, defaults to cwd
});
```

I did separate out `runPostBuildScript.ts` into its own script, but on the whole, no existing functionality is changed from a `keycloakify` bin standpoint.

## Summary by CodeRabbit

- **New Features**
  - Added a public API for generating Keycloak theme resources from completed build output.
  - Supports specifying a custom resources directory and returns its generated location.
  - Automatically runs supported post-build processing after resource generation.
  - Added a flat distributable API bundle for easier integration.

- **Bug Fixes**
  - Added safeguards to prevent unsafe directory deletion, invalid paths, incomplete builds, and recursive post-build execution.
  - Ensures generation fails clearly when no theme resources are produced.


garronej:
Hello @nrakochy,

I’m sorry it has taken me so long to respond to your issue and review this PR.

The concern you raised is valid, and I appreciate the thought you put into the implementation. Several organizations need to build Keycloak themes in fully JavaScript-based environments where Maven and Git may not be available, this is a valid need. 
However supporting external hooks would also add complexity to both the documentation and ongoing maintenance. Beleve me I have tried to do without Maven, make keycloakify more standalone, but I faced a cascade of issues and ended up back peddaling. 

To keep the project manageable, I need to limit the number of build options and branching paths. Otherwise, every change to the build process becomes significantly harder to reason about and carries a greater risk of breaking one of the supported configurations.

Your implementation looks good, and I encourage you to use it in your project through [[patch-package](https://www.npmjs.com/package/patch-package)](https://www.npmjs.com/package/patch-package). Unfortunately, I don’t have the bandwidth to merge and maintain it as part of Keycloakify.

I hope you understand, and thank you again for taking the time to contribute.

Best,

PR #1049, opened 2026-08-05
fix(account): don't drop application.client from the kcContext

### LINKED ISSUE #1040: Multi-Page account theme: `applications.applications[].client` is dropped from kcContext
Hi! First off, thanks so much for Keycloakify — it's been a joy to work with. 🙏

I ran into a small regression on the **Multi-Page account theme** and wanted to flag it.

**What happens**
On the `applications.ftl` page, `kcContext.applications.applications` is still there, but every entry is now missing its `client` object. So `application.client.clientId`, `.name`, etc. are all gone in the browser.

**What I found**
The kcContext serializer aborts any object whose class starts with `org.keycloak.models`:

```ftl
<#list ["models", "services", "authentication", ...] as namespacePortion>
    <#if stringified?starts_with("org.keycloak." + namespacePortion)>
        <#return abort>
    </#if>
</#list>
```

Each `application.client` is a `ClientModel` (`org.keycloak.models.*`), so it matches `"models"` and the whole object gets dropped. The Java bean is fine — it still returns `getClient()`.

**When it started**
- ✅ Works in `v11.8.0` (no `org.keycloak.models` blocklist yet)
- ❌ Broken in `v11.15.11` (blocklist added)

So it looks like the blocklist (great for not leaking internals!) and the Multi-Page `applications.ftl` page just conflict, since that page really needs to expose a `ClientModel`.

Hope this helps. Thanks again! 😊

e210:
On the `applications.ftl` page of the Multi-Page account theme, every `application.client` is a `ClientModel`, so the `org.keycloak.models` blocklist dropped it. The stock `Applications` page then throws on `application.client.clientId` and renders nothing.

Adds an exception scoped to that exact path, next to the existing `OTPPolicy` one. Verified end to end on Keycloak 26.7.0.

Fixes #1040

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where application data could fail to load on the applications page when it included client details.
  * Improved serialization of application information for more reliable page rendering.

### DIFF
--- a/src/bin/keycloakify/generateFtl/kcContextDeclarationTemplate.ftl
+++ b/src/bin/keycloakify/generateFtl/kcContextDeclarationTemplate.ftl
@@ -240,6 +240,19 @@ function decodeHtmlEntities(htmlStr){
                 <#if stringified?starts_with("org.keycloak.models.RequiredActionProviderModel") >
                     <#break>
                 </#if>
+                <#--
+                    On the applications.ftl page of the multi page account theme, each
+                    application.client is a ClientModel. It's not a Keycloak internal leaking
+                    into the kcContext, the page is meant to expose it.
+                    See: https://github.com/keycloakify/keycloakify/issues/1040
+                -->
+                <#if
+                    xKeycloakify.themeType == "account" &&
+                    xKeycloakify.pageId == "applications.ftl" &&
+                    areSamePath(path, [ "applications", "applications", "*", "client" ])
+                >
+                    <#break>
+                </#if>
                 <#list ["models", "services", "authentication", "quarkus.runtime", "transaction", "connections", "utils.ClosingStream"] as namespacePortion>
                     <#if stringified?starts_with("org.keycloak." + namespacePortion)>
                         <#return abort>


src/bin/keycloakify/generateFtl/kcContextDeclarationTemplate.ftl
```diff
@@ -240,6 +240,18 @@ function decodeHtmlEntities(htmlStr){
...
+                <#--
+                    On the applications.ftl page of the multi page account theme, each
+                    application.client is a ClientModel. It's not a Keycloak internal leaking
+                    into the kcContext, the page is meant to expose it.
+                    See: https://github.com/keycloakify/keycloakify/issues/1040
+                -->
+                <#if
+                    xKeycloakify.pageId == "applications.ftl" &&
+                    areSamePath(path, [ "applications", "applications", "*", "client" ])
+                >
```
garronej, line 253:
```suggestion
                <#if
                    xKeycloakify.themeType == "account" &&
                    xKeycloakify.pageId == "applications.ftl" &&
                    areSamePath(path, [ "applications", "applications", "*", "client" ])
                >
```

Maybe, just for the sake of transparency (because there's currently no "application.ftl" page in the login ui)

garronej (changes requested):
Thank you!

src/bin/keycloakify/generateFtl/kcContextDeclarationTemplate.ftl
e210: Applied, thanks. It also matches how the `index.ftl` rules are guarded further down, so it reads more consistently that way.

Double-checked the behaviour: with `themeType` set to `login` and the same pageId, the client goes back to being filtered.

PR #1053, opened 2026-08-12
fix: WebAuthn registration in latest Keycloak

### LINKED ISSUE #1050: WebAuthn passwordless broken on Keycloak 26.7
## Summary

Keycloak 26.7 replaced the WebAuthn policy field `requireResidentKey` with `residentKey` so  the `requireResidentKey` is deprecated.

Keycloakify has no knowledge of `residentKey` so it actively sends `authenticatorSelection.requireResidentKey = false` to `navigator.credentials.create()`, which instructs the authenticator **not** to create a discoverable credential.

Registration appears to succeed. Passwordless / usernameless passkey login then fails, because no discoverable credential exists. There is no error message at any point.

## Steps to reproduce

1. Deploy a Keycloakify login theme (`11.15.14`) to Keycloak `26.7`.
2. In the admin console, set **WebAuthn Passwordless Policy → Resident Key = `required`**.
3. Register a passkey (I use Chrome Webauthn devtools)
4. Attempt passwordless login with that passkey.

**Expected:** a discoverable (resident) credential is created; passwordless login works.

**Actual:** a non-discoverable credential is created (the field `Is resident` in credentials record is blank); passwordless login cannot find it. No error is surfaced.

## Root cause

Dropped in three layers:

1. in KcContext, type declares only `requireResidentKey: string`.
2. `WebauthnRegister.useScript.tsx` — the `input` object is a hand-written allowlist; fields not listed never reach the script.
3. `js/webauthnRegister.js` only branches on `input.requireResidentKey`, no `residentKey` handling.

On 26.7, if the server no longer populates the deprecated `requireResidentKey`, then `JSON.stringify(undefined)` returns `undefined`, so the generated script literally contains:

```js
requireResidentKey : undefined,
```

So Keycloakify explicitly instructs the authenticator to create a **non-discoverable** credential. This is worse than omitting the setting, and the failure is completely silent.Just update the latest script for fix won't be enough. We also need  backward compatible for older Keycloak version which never sends `residentKey` and keeps the legacy path; 26.7+ takes the new one

rolandspannagl:
I found two issues while running recent Keycloak with WebAuthn (login-ui):
- registerByWebAuthn() ignored PublicKeyCredential.authenticatorAttachment, so the registration request didn't include it.
- Only legacy requireResidentKey was handled in authenticatorSelection, so the new setting residentKey was ignored.

This is the companion fix for keycloakify/login-ui.

It also fixes https://github.com/keycloakify/keycloakify/issues/1050.

## Summary by CodeRabbit

* **Enhancements**
  * Improved WebAuthn registration by honoring the configured resident key requirement.
  * Registration requests now include the authenticator attachment setting.
  * WebAuthn registration configuration is consistently carried through the registration flow.
* **Bug Fixes**
  * Fixed inconsistencies between registration settings, form submissions, and WebAuthn requests.

### DIFF
--- a/src/login/KcContext/KcContext.ts
+++ b/src/login/KcContext/KcContext.ts
@@ -431,7 +431,7 @@ export declare namespace KcContext {
         attestationConveyancePreference: string;
         authenticatorAttachment: string;
         requireResidentKey: string;
-        residentKey: string;
+        residentKey?: string;
         userVerificationRequirement: string;
         createTimeout: number | string;
         excludeCredentialIds: string;
--- a/src/login/KcContext/KcContext.ts
+++ b/src/login/KcContext/KcContext.ts
@@ -431,6 +431,7 @@ export declare namespace KcContext {
         attestationConveyancePreference: string;
         authenticatorAttachment: string;
         requireResidentKey: string;
+        residentKey: string;
         userVerificationRequirement: string;
         createTimeout: number | string;
         excludeCredentialIds: string;
--- a/src/login/KcContext/kcContextMocks.ts
+++ b/src/login/KcContext/kcContextMocks.ts
@@ -540,6 +540,7 @@ export const kcContextMocks = [
         attestationConveyancePreference: "direct",
         authenticatorAttachment: "platform",
         requireResidentKey: "required",
+        residentKey: "required",
         userVerificationRequirement: "preferred",
         createTimeout: 60000,
         excludeCredentialIds: "credId123,credId456",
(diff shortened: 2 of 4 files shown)


garronej:
Thanks for the PR.  

It has been addressed in https://github.com/keycloakify/keycloakify/commit/0ede87bbf006baf1513df62f612b1ad8c0b587fd
and 
https://github.com/keycloakify/login-ui/commit/c86a6cc0a3aa6fa65e96b624c15177b8cabc4321  

New version of keycloakify has been released (11.16.0).

Best,

